PR #38, opened 2023-10-25
[PI-52] product team domain logic

nomad3k:


feature_tests/features/product_team.failure.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Product Teams Failure Scenarios
+
+  Scenario: Failure - invalid Product Team Name
+    Given ODS Organisations:
+      | ods_code | name    |
+      | H8S7A    | BURENDO |
+    And Users:
+      | user_id          | name      |
+      | test@example.org | Test User |
+    When User "test@example.org" creates Product Team {00702d39-e65f-49f5-b9ef-6570245bfe17} called " " for supplier H8S7A
```
jaklinger, line 10: few points:

1. seems like a inconsistent syntax: tabular first and then verbose second
2. it's not possible from this line to know that `test@example.org` is an ID (or is it a name or email?) and what `00702d39-e65f-49f5-b9ef-6570245bfe17` and " " correspond to.
3. Is supplier the right word? Who is the supplier of NRLF?

feature_tests/features/product_team.failure.feature
```diff
@@ -0,0 +1,32 @@
...
+    Then the operation was not successful
+    And the error is AssertionError
```
jaklinger, line 12: python assertion errors are generally not seen as something to catch

feature_tests/features/product_team.success.feature
```diff
@@ -0,0 +1,37 @@
...
+      | system   | OdsOrganisation |
+      | value    | <ods_code>      |
+    And the result owner is a Reference with:
+      | property | value     |
+      | system   | User      |
+      | value    | <user_id> |
+    And the following events were raised:
+      | event                   |
+      | ProductTeamCreatedEvent |
+    And event #1 is ProductTeamCreatedEvent with:
```
jaklinger, line 27: `#1` is being rendered as a comment in the linter, not sure if that's an issue!

feature_tests/steps/common.py
```diff
@@ -0,0 +1,91 @@
+from uuid import UUID
+
+import parse
+from behave import register_type
+
+
+@parse.with_pattern(
+    r"\{[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}\}"
```
jaklinger, line 8: i'd factor out that regex into a named constant - because it's not totally clear what it is (i mean i know its a uuid, but still)

feature_tests/steps/common.py
```diff
@@ -0,0 +1,91 @@
...
+    UUID=parse_uuid,
+    String=parse_string,
+    Number=parse_number,
+    OdsCode=parse_ods_code,
+    An=parse_a_or_an,
+    Path=parse_path,
+)
+
+
+def catch_errors(func):
```
jaklinger, line 49: alternatively a context manager would've been cool here, e.g.

```python
@contextmanager
def catch_errors(context):
    try: 
        yield
    except Exception as e:
        context.error = e

...

with catch_error(context):
   do_something()
```

feature_tests/steps/common.py
```diff
@@ -0,0 +1,91 @@
...
+            func(context, *args, **kwargs)
+        except Exception as e:
+            context.error = e
+
+    return inner
+
+
+def assert_type_matches(obj, expected):
+    actual = type(obj).__name__
+    assert actual == expected, "Type mismatch: expected {expected} got {actual}"
```
jaklinger, line 65: there's a slight chance of this giving false positives because using the name will circumvent namespacing.

Could add another parse method for types, so that the name is parsed into the correct type explicitly, and then you can instead check type equalities

feature_tests/steps/common.py
```diff
@@ -0,0 +1,91 @@
...
+        [head, *tail] = funcs
+        try:
+            return head(v)
+        except:
+            return __parse_value(v, *tail)
+
+    return __parse_value(v, parse_uuid, parse_number, parse_ods_code, parse_string)
+
+
+def read_value_from_path(obj, path) -> any:
```
jaklinger, line 83: i understand what this is doing, but that's only because i understand low level python (there are cases and object types that this definitely won't work for since __dict__ doesn't necessarily return attributes). At a minimum needs a docstring to explain that it's attempting to retrieve a value by json path in a python object's attributes

feature_tests/steps/event_steps.py
```diff
@@ -0,0 +1,35 @@
...
+    i = 0
+    for row in context.table:
+        assert i < len(context.events), "event count mismatch: too few raised"
+        actual = context.events[i]
+        expected = row["event"]
+        assert (
+            type(actual).__name__ == expected
+        ), "Event expected {expected} got {actual}"
+        i = i + 1
+    assert len(context.events) == i, "event count mismatch: too many raised"
```
jaklinger, line 21: pretty sure this can be compacted to:

```suggestion
    assert len(context.events) == len(context.table.rows), "event count mismatch: too many raised"
    
    for row, event in zip(context.table, context.events):
        expected = row["event"]
        assert (
            type(event).__name__ == expected
        ), "Event expected {expected} got {event}"
```

feature_tests/steps/ods_steps.py
```diff
@@ -0,0 +1,20 @@
+from behave import given
+from domain.core.error import DuplicateError
+from domain.core.root import Root
+
+
+def _add_ods_to_context(context, id: str, name: str):
+    if id in context.ods_organisations:
+        raise DuplicateError("Invalid test: ODS Organisation exists")
```
jaklinger, line 8: mmm i don't this logic belongs here. What if we were genuinely testing the effect of creating the same org twice? I think that still falls under the realm of domain logic

feature_tests/steps/ods_steps.py
```diff
@@ -0,0 +1,20 @@
+from behave import given
+from domain.core.error import DuplicateError
+from domain.core.root import Root
+
+
+def _add_ods_to_context(context, id: str, name: str):
```
jaklinger, line 6: i know it's shorter for brevity, but ods can mean code or org, and i guess we mean, so  `_add_ods_org_to_context` would be easier to follow

feature_tests/steps/operation_steps.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+@then("the operation was successful")
+def step_impl(context):
+    assert context.error is None, f"Unexpected error: {context.error}"
+
+
+@then("the operation was not successful")
+def step_impl(context):
+    assert context.error, f"Expected error"
```
jaklinger, line 11: think the error message is missing something

feature_tests/steps/product_team_steps.py
```diff
@@ -0,0 +1,20 @@
...
+    'User "{user_id}" creates Product Team {id:UUID} called "{name:String}" for supplier {ods_code:OdsCode}'
+)
+@catch_errors
+def step_impl(context, user_id, id, name, ods_code):
+    if user_id not in context.users:
+        raise NotFoundError("Unknown User")
+    if ods_code not in context.ods_organisations:
+        raise NotFoundError("Unknown ODS Organisation")
+    owner = context.users[user_id]
+    context.subject = context.ods_organisations[ods_code]
```
jaklinger, line 17: if this statement was run twice this would be overwritten, which would lead to strange behaviour that I wouldn't expect the test team to follow

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
+from behave import then
+from domain.core.reference import Reference
+
+from feature_tests.steps.common import parse_value, read_value_from_path
+
+
+@then("the result is of type {type_name}")
+def step_impl(context, type_name):
+    t = type(context.result).__name__
+    assert t == type_name, f"Expected '{type_name}' got '{t}'"
```
jaklinger, line 10: i thought there was already a function you wrote for doing this

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@then("the result {field} equals {value:Number}")
+def step_impl(context, field, value):
+    assert context.result is not None, "result"
+    v = context.result.__dict__[field]
```
jaklinger, line 16: more generally the following is preferred, since __dict__ doesn't always return attributes:

```suggestion
    v = getattr(result, field)
```

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+
+@then("the result is of type {type_name}")
+def step_impl(context, type_name):
+    t = type(context.result).__name__
+    assert t == type_name, f"Expected '{type_name}' got '{t}'"
+
+
+@then("the result {field} equals {value:Number}")
+def step_impl(context, field, value):
+    assert context.result is not None, "result"
```
jaklinger, line 15: the error message here is just the word `"result"`

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+    v = context.result.__dict__[field]
+    assert v == value, f"Expected {value} but found {v}"
+
+
+@then('the result {field:String} equals "{value:String}"')
+def step_impl(context, field, value):
+    assert context.result is not None, "result"
+    v = context.result.__dict__[field]
+    assert v == value, f"Expected {value} but found {v}"
```
jaklinger, line 24: given that these steps are all the same couldn't we just do away with the type parsing?

jaklinger: seems like a waste to rewrite them

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@then('the result {field:Path} is a Reference to {system:String} "{value:String}"')
+def step_impl(context, field, system, value):
+    ref = context.result.__dict__[field]
+    assert isinstance(ref, Reference), "Is reference"
```
jaklinger, line 37: error message doesn't mean anything

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+    assert ref.system == system, f"Expected '{system}' got '{ref.system}'"
+    assert ref.value == value, f"Expected '{value}' got '{ref.value}'"
+
+
+def _assert_value(obj, path, value):
```
jaklinger, line 42: should probably be calling the other similar function for accessing path objects, to avoid duplication. Otherwise, same comment regarding a docstring

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+    if len(tail) > 0:
+        _assert_value(v, tail, value)
+    else:
+        assert (
+            v == value
+        ), f"Expected '{type(value).__name__}:{value}' got '{type(v).__name__}:{v}'"
+
+
+@then("the result is")
+def step_impl(context):
```
jaklinger, line 54: this needs a doc string at the bare minimum - it's not easy to draw a line between what's happening here and what's in the table

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+    for row in context.table:
+        _assert_value(context.result, row["path"].split("."), parse_value(row["value"]))
+
+
+@then("the result is {a:An} {type_name:String} with")
+def step_impl(context, a, type_name):
+    assert context.result, "Expected result"
```
jaklinger, line 61: error message isn't saying anything

feature_tests/steps/result_steps.py
```diff
@@ -0,0 +1,87 @@
...
+        expected_value = parse_value(row["value"])
+        actual_value = read_value_from_path(value, property_name)
+        assert (
+            actual_value == expected_value
+        ), f"Expected value '{expected_value}' got '{actual_value}'"
+
+
+@then("the result {path:Path} is {a:An} {type_name:String} with")
+def step_impl(context, path, a, type_name):
+    assert context.result, "Expected result"
```
jaklinger, line 76: error message isn't saying anything

src/layers/domain/core/ods_organisation.py
```diff
@@ -0,0 +1,35 @@
...
+        self,
+        id: str,
+        name: str,
+    ):
+        assert_is_ods_code(id)
+        super().__init__(id, name)
+
+    def create_product_team(
+        self, id: str, name: str, owner: User
+    ) -> (ProductTeam, ProductTeamCreatedEvent):
```
jaklinger, line 26:
```suggestion
    ) -> tuple[ProductTeam, ProductTeamCreatedEvent]:
```

src/layers/domain/core/ods_organisation.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        Creates an instance of a ProductTeam
+        """
+        assert isinstance(owner, User), "owner is required"
```
jaklinger, line 30: assertions are frowned upon outside of tests

jameslinnell:
Could you rebase please

src/layers/domain/core/product.py
```diff
@@ -1,16 +1,20 @@
...
+    A Product represents logical and physical software products.  A Product may
+    be referenced by multiple keys, such as internal GUID, Product Id, ASID,
+    etc.
+    """
 
-
-class Product(Entity, QuestionnaireEntity):
     def __init__(
         self,
         id: str,
```
jaklinger, line 20: guess this should be `UUID` give the above generic[UUID] ?

```suggestion
        id: UUID,
```

src/layers/domain/core/questionnaire.py
```diff
@@ -33,11 +49,26 @@ def add_question(
...
+        locked this method will throw an error.
+        """
         raise NotImplementedError()
+
+    def revise(self) -> "Questionnaire":
+        """
+        Creates an exact duplicate of the questionnaire, which is the only way
+        to edit questionnaires once they become locked.
+        """
+        raise NotImplemented()
```
jaklinger, line 74:
```suggestion
        raise NotImplementedError()
```

src/layers/domain/core/entity.py
```diff
@@ -18,3 +23,6 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return not self.__eq__(other)
+
+    def as_reference(self) -> Reference:
```
jaklinger, line 27: the whole `Reference` thing comes across as a code smell: why are converting objects to string references rather referencing the actual objects? Python by default will treat objects as references, and if the issue is about namespacing ids then we can just modify the `__hash__` function above to include the type.__name__

src/layers/domain/core/reference.py
```diff
@@ -0,0 +1,30 @@
+class Reference:
```
jaklinger, line 1: See comment above: basically - the reference class reads as a code smell when we've already introduced hashing in the entity class and objects are anyway treated as references in python

src/layers/domain/core/reference.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    A class used for storing references to other classes
+    """
+
+    def __init__(self, system, value, name):
+        self.system = system
+        self.value = str(value)
```
jaklinger, line 8: i wonder if this casting will have unexpected side effects

src/layers/domain/core/reference.py
```diff
@@ -0,0 +1,30 @@
...
+            and self.system == other.system
+            and self.value == other.value
+        )
+
+    def __ne__(self, other):
+        """
+        Reference inequality is based on the system/value.  The name is just
+        decorative.
+        """
+        return not self.__eq__(other)
```
jaklinger, line 30: this is already the default behaviour

```suggestion
```

src/layers/domain/core/tests/test_ods_organisation.py
```diff
@@ -0,0 +1,31 @@
...
+        ["RTG09", "QUEENS MEDICAL CENTRE"],
+        ["PN103", "THE ROYAL OLDHAM HOSPITAL"],
+    ],
+)
+def test__can_instantiate_ods_organisation(id: str, name: str):
+    result = Root.create_ods_organisation(id, name)
+
+    assert isinstance(result, OdsOrganisation), "OdsOrganisation instantiated"
+    assert result.id == id, "Id set"
+    assert result.name == name, "Name set"
```
jaklinger, line 19: error messages aren't useful

src/layers/domain/core/tests/test_product_team.py
```diff
@@ -0,0 +1,34 @@
...
+    assert result.name == name, "name mismatch"
+    assert result.organisation == subject.as_reference(), "organisation mismatch"
+    assert result.owner == user.as_reference(), "owner mismatch"
+
+    assert isinstance(event, ProductTeamCreatedEvent), "Event type mismatch"
+    assert event.id == id, "event id mismatch"
+    assert event.id == id, "event id mismatch"
+    assert event.name == name, "event name mismatch"
+    assert event.organisation == subject.as_reference(), "event organisation mismatch"
+    assert event.owner == user.as_reference(), "event owner mismatch"
```
jaklinger, line 34: error messages aren't useful

feature_tests/steps/product_team_steps.py
```diff
@@ -0,0 +1,20 @@
...
+
+@when(
+    'User "{user_id}" creates Product Team {id:UUID} called "{name:String}" for supplier {ods_code:OdsCode}'
+)
+@catch_errors
+def step_impl(context, user_id, id, name, ods_code):
+    if user_id not in context.users:
+        raise NotFoundError("Unknown User")
+    if ods_code not in context.ods_organisations:
+        raise NotFoundError("Unknown ODS Organisation")
```
jaklinger, line 15: really shouldn't be raising and catching these errors in our tests since it's not actually a part of the domain logic

jaklinger: I guess this is a test configuration error, rather than a domain error, so I'm suggesting `raise TestConfigurationError` so we can filter these out in the assertions

feature_tests/features/product_team.failure.feature
```diff
@@ -0,0 +1,32 @@
...
+
+  Scenario: Failure - unknown supplier
```
jaklinger, line 14: This isn't testing domain logic -it's testing the test configuration (i.e. no domain code is called to create a product team)

jaklinger:
superseded by #38

PR #46, opened 2023-10-27
[feature/PI-72-validate_ods_code] Validate ODS Code

jaklinger:


.github/workflows/pull-requests.yml
```diff
@@ -79,13 +79,16 @@ jobs:
   test--unit:
     needs: [build-head]
     runs-on: [self-hosted, ci]
+    strategy:
+      matrix:
+        test-type: [unit, slow]
```
Rohoolio, line 84: This looks like a sneaky extra change haha - what does this achieve?

src/layers/domain/ods/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    "https://directory.spineservices.nhs.uk/ORD/2-0-0/organisations/{ods_code}"
+)
+BACKOFF_BASE_SECONDS = 2
+
+
+class UnexpectedStatusCode(Exception):
+    pass
+
+
+class OdsApiOffline(ExceptionGroup):
```
Rohoolio, line 19: Are we sure the only way we can get errors is if the ODS API is offline? at first glance I assumed it had an offline version of the data to hit reading this - do we want something of a naming convention for the exception groups to make it more obvious or would that be overthinking it?

src/layers/domain/ods/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+def is_valid_ods_code(ods_code: str) -> bool:
+    url = _construct_ods_url(ods_code=ods_code)
+    response = requests.get(url=url)
+    status_code = HTTPStatus(response.status_code)
+
+    if status_code is HTTPStatus.OK:
+        return True
+    if HTTPStatus.BAD_REQUEST <= status_code < HTTPStatus.INTERNAL_SERVER_ERROR:
+        return False
+    if status_code >= HTTPStatus.INTERNAL_SERVER_ERROR:
```
Rohoolio, line 73: Is there no way for the Exception group to handle these, especially if they are expected codes?

jaklinger: these aren't exceptions, they're enums

jaklinger: so basically its the same as:

src/layers/domain/ods/__init__.py
jaklinger: > Are we sure the only way we can get errors is if the ODS API is offline? 
Yes - any other error will lead us to raise our own generic internal server error, and `OdsApiOffline` can only be raised if there are 3 consecutive internal server errors on the ODS API

> at first glance I assumed it had an offline version of the data to hit reading this - do we want something of a naming convention for the exception groups to make it more obvious or would that be overthinking it?
could rename to `OdsApiOfflineError`?

.github/workflows/pull-requests.yml
jaklinger: not sure i follow - otherwise the `slow` tests i wrote won't run?

src/layers/domain/ods/__init__.py
Rohoolio: ok just me being dense then ha

Rohoolio (approved):

PR #63, opened 2023-11-10
[PI-119] FHIR/Swagger/Pydantic

jaklinger:


.github/actions/components/caching/action.yml
```diff
@@ -22,6 +22,8 @@ runs:
           **/tfplan
           **/swagger.yaml
           **/infrastructure/terraform/**/output.json
+          .downloads
```
Rohoolio, line 25: This relates to all the downloady bits, im sure we didnt do it this way with NRLF so wondering whats different this time for pulling down the JAR file? The only thing I can think of is that we are doing this on the fly in the CI?

.tool-versions
```diff
@@ -4,3 +4,4 @@ python 3.11.4
 terraform 1.5.7
 jq 1.7
 yq 4.34.2
+nodejs 20.9.0
```
Rohoolio, line 7: Is this now needed on the packer image? Starting to sound like that ticket could do with being brought in soon haha

.github/actions/components/caching/action.yml
jaklinger: i just needed somewhere to download the JAR file to :D don't mind where it goes, but somewhere at the repo root that is gitignored. What do you suggest?

.tool-versions
jaklinger: ah yeah i'll add that to the ticket

PR #79, opened 2023-11-22
[feature/PI-118-product_team_e2e] feature tests and rejig response matrix

jaklinger:


src/api/readProductTeam/src/v1/tests/test_steps.py
```diff
@@ -1,14 +1,14 @@
-from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
-from hypothesis import given
-from hypothesis.strategies import builds
+# from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
```
jameslinnell, line 1: Commented code?

jameslinnell (approved):

src/api/readProductTeam/src/v1/tests/test_steps.py
jaklinger: it's not totally clear to me whether or not we need this test, but will wait til PI-124 until i delete the file

PR #80, opened 2023-11-22
[feature/PI-110-step_chain] Refactor step chain

jaklinger:


src/layers/event/api_step_chain/__init__.py
```diff
@@ -0,0 +1,33 @@
+from types import ModuleType
+
+from event.logging.step_decorators import logging_step_decorators
+from event.response.steps import response_steps
+from event.step_chain import StepChain
+from event.versioning.constants import VERSIONING_STEP_ARGS
+from event.versioning.steps import get_steps_for_requested_version, versioning_steps
+
+STEP_DECOS = [*logging_step_decorators]
```
Rohoolio, line 9: Little colloquial here? 😂

jaklinger: it made the rest of a file a bit easier to read than when it forced new lines everywhere

src/api/createProductTeam/index.py
```diff
@@ -1,38 +1,22 @@
...
 
 class Environment(BaseEnvironment):
     DYNAMODB_TABLE: str
 
 
 cache = {**Environment.build().dict()}
-step_decorators = [*logging_step_decorators]
-pre_steps = [*versioning_steps, *event_processing_steps]
-post_steps = [*response_steps]
+versioned_steps = {"1": v1_steps}
```
Rohoolio, line 13: Just a thought for keeping things safe, this versioned_steps as a constant in the class doesnt enforce this structure, might be worth having a class that requires a version and steps to ensure its a forced pattern?

src/layers/event/versioning/constants.py
```diff
@@ -2,5 +2,8 @@
 
 VERSION_HEADER_PATTERN = r"^(\d+)$"
 VERSION_RE = re.compile(r"^v(\d+)$")
-API_ROOT_DIRNAME = "api"
-VERSIONED_HANDLER_GLOB = "src/v*/steps.py"
+
+
+class VERSIONING_STEP_ARGS:
+    VERSIONED_STEPS = "versioned_steps"
```
Rohoolio, line 8: Related to the comment above where this default doesnt match the object of {"1": v1_steps}

Rohoolio: I know you type enforce it in the functions, just a suggestion to make that a bit more secure

Rohoolio (approved):

src/layers/event/api_step_chain/__init__.py
jaklinger: fine done :)

PR #85, opened 2023-11-24
[feature/PI-134-developers_to_use_developer_role] Use developer role

jameslinnell:


infrastructure/terraform/per_workspace/locals.tf
```diff
@@ -2,4 +2,5 @@ locals {
   region       = "eu-west-2"
   project      = "nhse-cpm"
   current_time = timestamp()
+  ws_type      = replace(terraform.workspace, "_", "-") == "dev" || replace(terraform.workspace, "_", "-") == "ref" || replace(terraform.workspace, "_", "-") == "int" || replace(terraform.workspace, "_", "-") == "prod" || replace(terraform.workspace, "_", "-") == "dev-sandbox" || replace(terraform.workspace, "_", "-") == "int-sandbox" || replace(terraform.workspace, "_", "-") == "ref-sandbox" ? "${var.workspace_type} ${upper(replace(terraform.workspace, "_", "-"))}" : "${var.workspace_type}"
```
Rohoolio, line 5: Minor but might as well call it workspace_type

infrastructure/terraform/per_workspace/outputs.tf
```diff
@@ -2,6 +2,10 @@ output "dynamodb_table_name" {
   value = module.table.dynamodb_table_name
 }
 
+# output "assumed_role" {
```
Rohoolio, line 5: Commented code?

scripts/infrastructure/roles/manage-non-mgmt-aws-policies.sh
```diff
@@ -0,0 +1,115 @@
...
+        fi
+      else
+        aws iam attach-role-policy \
+          --role-name "${role}" \
+          --policy-arn "arn:aws:iam::${ACCOUNT_ID}:policy/${policy_name}" \
+          --region "${AWS_REGION_NAME}" \
+          || exit 1
+      fi
+    done
+  else
```
Rohoolio, line 52: Does this actually update the policy? From what I can tell this section runs if the policy does exist but the code seems to attach that role again but I cant tell if it has redeployed the new one?

Rohoolio: Or does updating the version of the policy later update it for all Roles that are using it?

scripts/infrastructure/roles/manage-non-mgmt-aws-roles.sh
```diff
@@ -0,0 +1,74 @@
...
+        --region "${AWS_REGION_NAME}" \
+        || exit 1
+    fi
+
+    aws iam get-role --role-name "NHSIntegrationRole" &> /dev/null
+    if [ $? != 0 ]; then
+      tf_assume_role_policy=$(_substitute_environment_variables ./scripts/infrastructure/policies/role-trust-policy.json)
+
+      aws iam create-role \
+        --role-name "NHSIntegrationRole" \
```
Rohoolio, line 68: I think considering this is a Role that will only be used by CI for the sake of testing I think the name should reflect that - some suggestions:
- NHSTestCIRole
- NHSIntegrationTestCIRole
- NHSCIIntegrationTestRole
- NHSIntegrationTestRole_CI

Tbh they are all a bit rubbish haha

nomad3k (changes requested):

scripts/infrastructure/roles/manage-non-mgmt-aws-roles.sh
jameslinnell: Changed to NHSTestCIRole

PR #90, opened 2023-11-29
Release/2023-11-29-a

jaklinger:


.github/workflows/pull-requests.yml
```diff
@@ -197,6 +197,26 @@ jobs:
           test-type: integration
           requires-aws: true
 
+  test--feature--integration:
+    needs:
```
jameslinnell, line 201: I know it's a bit late for this as it's already a release, but...
It only needs test--integration. Also test--integration only needs terraform-head-build
```suggestion
    needs:
    [
      test--integration
    ]
```

PR #93, opened 2023-12-04
[NRLF-127] Change Product to Device

nomad3k:
Renames "Product" to "Device", but also included some work on the Device domain model to fill in extra time.


src/layers/domain/core/accredited_system_id.py
```diff
@@ -0,0 +1,22 @@
+import re
+
+ACCREDITED_SYSTEM_ID = "^[0-9]{1,12}$"
```
jaklinger, line 3: duplication of a field thats in `validation.py` in this folder

src/layers/domain/core/device.py
```diff
@@ -0,0 +1,351 @@
...
+    InvalidDeviceKeyError,
+    InvalidKeyError,
+    NotFoundError,
+)
+from domain.core.event import Event
+from domain.core.product_id import ProductId
+from pydantic import BaseModel, Field
+
+
+class DeviceCreatedEvent(Event):
```
jaklinger, line 18: I've rewritten these as native python dataclasses to avoid hand rolled data missing fields on serialising to dynamodb

src/layers/domain/core/device.py
```diff
@@ -0,0 +1,351 @@
...
+        del self.relationships[target_id]
+        event = RelationshipRemovedEvent(id=self.id, target=target_id)
+        return self.add_event(event=event)
+
+    def add_key(self, key: ProductId | AccreditedSystemId) -> DeviceKeyAddedEvent:
+        """
+        Adds a new Key to a Device
+        """
+        if key in self.keys:
+            raise DuplicateError(f"Key {key} exists")
```
jaklinger, line 312: the error message here should be:

```suggestion
            raise DuplicateError(f"Key '{key}' has been provided multiple times.")
```

as there is no concept or persistence in this layer

src/layers/domain/core/device.py
```diff
@@ -0,0 +1,351 @@
...
+        if isinstance(key, ProductId):
+            type = DeviceKeyType.PRODUCT_ID
+        elif isinstance(key, AccreditedSystemId):
+            type = DeviceKeyType.ACCREDITED_SYSTEM_ID
+        else:
+            raise InvalidDeviceKeyError(
+                "key must be ProductId() or AccreditedSystemId()"
+            )
```
jaklinger, line 320: for extensibility i'd have each key know it's own `DeviceKeyType`

src/layers/domain/core/device.py
```diff
@@ -0,0 +1,351 @@
...
+        self.keys[key] = DeviceKey(type=type)
+        event = DeviceKeyAddedEvent(id=self.id, key=key, type=type)
```
jaklinger, line 322: to avoid data integrity issues you need to do:

```suggestion
        self.keys[key] = DeviceKey(type=type)
        event = DeviceKeyAddedEvent(id=self.id, **self.keys[key].dict())
```

src/layers/domain/core/error.py
```diff
@@ -4,3 +4,11 @@ class DuplicateError(Exception):
 
 class NotFoundError(Exception):
     pass
+
+
+class InvalidDeviceKeyError(Exception):
+    pass
+
+
+class InvalidKeyError(Exception):
```
jaklinger, line 13: not very specific?

src/layers/domain/core/ods_code.py
```diff
@@ -0,0 +1,14 @@
+import re
```
jaklinger, line 1: i thought we agreed that we don't need to validate ods codes any more, as the domain can assume that ods codes are valid?

src/layers/domain/core/product_id.py
```diff
@@ -0,0 +1,48 @@
...
+2. similar characters have been removed from the set of available characters.
+  e.g. 0/O/Q, 1/I, S/5, B/8, and Z/2
+
+This leaves 26+10-11 = 25 unique characters
+9 characters per id gives 3.8 billion unique values: 25^9 = 3,814,697,265,625
+"""
+
+PRODUCT_ID_CHARS = "ACDEFGHJKLMNPRTUVWXY34679"
+PRODUCT_ID_REGEX = (
+    f"^[{PRODUCT_ID_CHARS}]{{3}}-[f{PRODUCT_ID_CHARS}]{{3}}-[{PRODUCT_ID_CHARS}]{{3}}$"
```
jaklinger, line 19: need to use raw strings for regex

```suggestion
    rf"^[{PRODUCT_ID_CHARS}]{{3}}-[f{PRODUCT_ID_CHARS}]{{3}}-[{PRODUCT_ID_CHARS}]{{3}}$"
```

src/layers/domain/core/product_id.py
```diff
@@ -0,0 +1,48 @@
...
+            raise InvalidProductIdError(f"Invalid Product Id: {content}")
+        return str.__new__(cls, content)
+
+    @staticmethod
+    def generate(seed: int | None = None) -> "ProductId":
+        high = len(PRODUCT_ID_CHARS) - 1
+        random.seed(seed)
+        return "-".join(
+            [
+                "".join([PRODUCT_ID_CHARS[random.randint(0, high)] for _ in range(3)])
```
jaklinger, line 45:
```suggestion
                "".join(random.choices(PRODUCT_ID_CHARS), k=3))
```

src/layers/domain/core/product_id.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __new__(cls, content):
+        if not re.match(PRODUCT_ID_REGEX, content):
+            raise InvalidProductIdError(f"Invalid Product Id: {content}")
+        return str.__new__(cls, content)
+
+    @staticmethod
+    def generate(seed: int | None = None) -> "ProductId":
+        high = len(PRODUCT_ID_CHARS) - 1
+        random.seed(seed)
```
jaklinger, line 42: given that this changes global state, better to do this externally (presume it's for testing consistency?)

src/layers/repository/device_repo.py
```diff
@@ -0,0 +1,255 @@
...
+                    {
+                        "pk": device_pk(event.id),
+                        "sk": device_pk(event.id),
+                        "pk_1": ods_pk(event.ods_code),
+                        "sk_1": device_pk(event.id),
+                        "name": event.name,
+                        "type": event.type,
+                        "ods_code": event.ods_code,
+                        "product_team_id": event.product_team_id,
+                        "status": event.status,
```
jaklinger, line 50: after switching to dataclasses you can do `**asdict(event)` here to avoid data integrity issues when we extend or change models (missing data)

src/layers/repository/keys.py
```diff
@@ -3,26 +3,31 @@
```
jaklinger, line 3: i was thinking to change this to the following, to make the repo easier to follow, without some magic string and making the pk/sk and associated methods more explicit:

```python
from enum import StrEnum
from typing import Generator

KEY_SEPARATOR = "#"


class TableKeys(StrEnum):
    DEVICE = "D"
    DEVICE_KEY = "DK"
    DEVICE_RELATIONSHIP = "DR"
    PRODUCT_TEAM = "PT"
    ODS_ORGANISATION = "O"

    def key(self, *args) -> str:
        return KEY_SEPARATOR.join((self, *args))

    def filter(
        self, iterable: list[dict[str, str]], key: str
    ) -> Generator[dict[str, str], None, None]:
        return (v for v in iterable if v[key].startswith(f"{self}{KEY_SEPARATOR}"))


def strip_key_prefix(key: str):
    _, tail = key.split(KEY_SEPARATOR, 1)
    return tail
```

src/layers/domain/core/device.py
```diff
@@ -0,0 +1,351 @@
...
+        TODO: Do we delete them or mark them inactive?
+        """
+        target_id = target.id if isinstance(target, Device) else target
+        if target_id not in self.relationships:
+            raise NotFoundError(f"Relationship {target_id} is unknown")
+        del self.relationships[target_id]
+        event = RelationshipRemovedEvent(id=self.id, target=target_id)
+        return self.add_event(event=event)
+
+    def add_key(self, key: ProductId | AccreditedSystemId) -> DeviceKeyAddedEvent:
```
jaklinger, line 307: this is missing some logic here on how the domain will decide where a key is a product id or asid - if we push this into the fhir layer then that will be domain logic spread up one tier, i'd have the function like:

```suggestion
    def add_key(self, key: str, type: str) -> DeviceKeyAddedEvent:
        # convert to product id or asid here
```

src/layers/repository/utils.py
```diff
@@ -7,12 +6,10 @@
...
     if value is None:
         return {"Null": True}
-    if isinstance(value, StrEnum):
-        return {"S": str(value)}
-    if isinstance(value, bool):  # isinstance(True,int) == True
+    if type(value) == bool:  # isinstance(True,int) == True
         return {"B": value}
-    if isinstance(value, int) or isinstance(value, float):
-        return {"N": value}
+    if type(value) == int or isinstance(value, float):
```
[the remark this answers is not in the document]
jaklinger: you don't need to commit the suggestion - but it is quite handy for side-by-side hints

src/layers/domain/core/product_id.py
nomad3k: changed to `rng = random.Random(seed)`

src/layers/repository/device_repo.py
nomad3k: ok, pick this up in your ticket.

src/layers/domain/core/device.py
nomad3k: The relationship is not bidirectional.  DVK relies on PID/ASID, but not the other way round.

jaklinger: yes exactly - i'm suggesting key type can be a property of the dvk that way only keys with types can be used

jaklinger (approved):

PR #113, opened 2023-12-21
[PI 109] custom attribute forwarding

Rohoolio:
* Changed made to the swagger to represent the changes needed in proxygen

### DIFF
--- a/infrastructure/swagger/04_apim.yaml
+++ b/infrastructure/swagger/04_apim.yaml
@@ -12,4 +12,17 @@ x-nhsd-apim:
     security:
       type: apikey
       header: apikey
-      secret: rowan-cpm-test-5
+      secret: rowan-cpm-test-8
+  target-attributes:
+    - name: nrl-ods-RJ11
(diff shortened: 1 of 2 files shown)


infrastructure/swagger/04_apim.yaml
```diff
@@ -12,4 +12,7 @@ x-nhsd-apim:
     security:
       type: apikey
       header: apikey
-      secret: rowan-cpm-test-5
+      secret: rowan-cpm-test-8
```
nomad3k, line 15: We can't push secrets into github.  We need to figure out how to handle this.

nomad3k (changes requested):

infrastructure/swagger/04_apim.yaml
Rohoolio: Thats the name of the secret and not the secret itself - and its stored within APIMs setup

jameslinnell (approved):

PR #122, opened 2024-01-10
[feature/PI-122-proxies] Add proxygen CI

jaklinger:


infrastructure/apigee/ptl/.proxygen/credentials.yaml
```diff
@@ -0,0 +1,4 @@
+client_id: connecting-party-manager-client # must match "api" in settings.yaml
```
Rohoolio, line 1: Im wondering if the concept of PTL even exists in Apigee - it sounds more like we just have prod credentials, although i accept that we already put our request in with that key_id - might be worth changing it at this stage to avoid confusion?

jaklinger: my impression was that there were three different sets if environments / workspaces:

* regular apim environments (ptl --> [internal-dev, internal-qa, etc], prod --> [dev, qa, int, prod], or something like that)
* proxygen environments (ptl --> [proxygen developers], prod -> [everyone else])
* cpm (dev --> [dev, local dev workspaces], ..., prod -> [prod])

and the "ptl" I've used corresponds to the regular apim environment, not the ptl one

Rohoolio: So are we going to need to create a second of credentials to manage the "prod" versions of regular APIM to CPM? I suppose thats actually a sensible idea even though we dont technically need two sets as a separation of concerns

infrastructure/swagger/04_apigee.yaml
```diff
@@ -1,15 +1,16 @@
 ---
 openapi: 3.0.3
 x-nhsd-apim:
+  temporary: ${ terraform_output("workspace_type.value") != "PERSISTENT" }
```
Rohoolio, line 4: 👍

scripts/infrastructure/apigee/proxygen.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    original_environ = os.environ.copy()
+    os.environ["HOME"] = os.environ["DOT_PROXYGEN"]
+    yield
+    os.environ["HOME"] = original_environ["HOME"]
+
+
+if __name__ == "__main__":
+    with temporary_home():
+        from proxygen_cli.cli.command_main import main
```
Rohoolio, line 21: When does the temporary home disappear? Just when the lambda dies?

PR #127, opened 2024-01-25
Feature/pi 187 questionnaire domain logic

megan-bower4:


src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
...
+
+class QuestionnaireResponse(BaseModel):
+    """
+    Represents a Questionnaire response mapping Questions to Responses.
+    """
+
+    responses: dict
+
+
+class QuestionnaireResponseValidator(BaseModel):
```
megan-bower4, line 87: QuestionnaireResponse

src/layers/domain/core/tests/test_questionnaire.py
```diff
@@ -1,80 +1,455 @@
...
+
+
+@pytest.mark.parametrize("name", ["alpha", "beta", "gamma"])
+def test_has_question(name: str):
+    subject = Questionnaire(name="questionnaire", version=1)
+    subject.add_question(name=name)
+
+    result = subject.__contains__(name)
+
+    assert result == True, "has_question"
```
megan-bower4, line 100: remove

src/layers/domain/core/tests/test_questionnaire.py
```diff
@@ -1,80 +1,455 @@
...
+        result2 = QuestionnaireResponseValidator(
+            questionnaire=questionnaire, responses=response
+        ).validate_and_get_invalid_responses()
+
+
+# def test_valid_questionnaire_response_rules():
+# not got any rules yet?
+
+# def test_invalid_questionnaire_response_rules():
+
```
megan-bower4, line 313: remove comment

src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
...
+        Adds a new question to the questionnaire.
         """
-        if name in self:
-            raise DuplicateError(f"Question exists: {name}")
-        result = Question(name=name, type=type, multiple=multiple)
-        self._questions.append(result)
-        return result
+        for question in self.questions:
+            if name == question.name:
+                raise DuplicateError(f"Question exists: {name}")
```
jaklinger, line 66: to reiterate - if using `dict[str, Question]` above then this becomes

```
if name in self:
    raise ...
```

src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
...
+    responses: dict
+
+
+class QuestionnaireResponseValidator(BaseModel):
+    """
+    Validates questionnaire responses against questionnaire questions are of the correct type, meet any validation criteria and .
+    """
+
+    questionnaire: Questionnaire
+    responses: dict
```
jaklinger: if we group all responses by question at the start then we can reduce the code later on, so I'd suggest `responses: list[tuple[str, list]]` (i.e. a list of question -> answers) for two reasons

* if we make the `responses` object a `list` then we can leverage pydantics `@validator(..., each_item=True)` to validate each response separately
* if we make all question answers `list`, even if multiple=False - this will allow us to simplify all downstream code as we won't need the `list`/not `list` branching

src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
...
+
+        return question_name in (q.name for q in self.questions)
 
     def add_question(
         self,
         name: str,
         type: QuestionType = QuestionType.STRING,
         multiple: bool = False,
+        validation_rules: list[str] = [],
+        choices: list[str] = [],
```
jaklinger, line 59: change to `set` as we don't want duplicates and `set` will offer us some nicer performance

src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
     name: str
     type: QuestionType
     multiple: bool
+    validation_rules: list[str] = []
+    choices: list[str] = []
```
jaklinger, line 34: change to set as we don't want duplicates and set will offer us some nicer performance

src/layers/domain/core/questionnaire.py
```diff
@@ -27,49 +30,238 @@ class Question(BaseModel):
...
+            invalid_responses_given = []
+            for q.name, self.responses[q.name] in invalid_responses:
+                invalid_response = f"Question '{q.name}' does not allow multiple responses. Response given: {self.responses[q.name]}."
+                invalid_responses_given.append(invalid_response)
+
+            raise InvalidResponseError(invalid_responses_given)
+
+        return True
+
+    def validate_and_get_invalid_responses(self):
```
jaklinger, line 227: so this is the function where the `@validator("responses", each_item=True)` will go, and then since we can then let pydantic do the error handling for us, simplify the entire function (with the other suggestions) to

```python
    @validator("responses", each_item=True))
    def validate_responses(self, response, values: dict[str, Questionnaire]):
        questionnaire = values.get("questionnaire")
        question_name, answers = response

        question = questionnaire.get(question_name)
        if question is None:
                raise InvalidQuestionResponse("question doesnt exist")    

        if not question.multiple and len(answers) > 1:
                raise InvalidQuestionResponse("too many")    

        for answer in answers:
            if not isinstance(answer, question.type):
                raise InvalidQuestionResponse("bad type")

            for validation_rule in validation_rules:
                validation_rule(answer)  # should raise a value error

            if question.choices and answer not in question.choices:
                raise InvalidQuestionResponse("bad choice")
```

src/layers/domain/core/questionnaire.py
```diff
@@ -1,75 +1,149 @@
...
         """
-        return question_name in (q.name for q in self._questions)
+
+        return question_name in self.questions
 
     def add_question(
         self,
         name: str,
-        type: QuestionType = QuestionType.STRING,
+        type: Type = str,
```
jaklinger, line 59: so a potential bug that might come up is that the word `type` also refers a python global called `type` (i.e. you can do `type("hello")` which returns `str`). It's actually quite an important object in python, so it might be worth renaming anything in this file called `type` to `answer_type` to avoid weird side-effects coming up at a later date

src/layers/domain/core/questionnaire.py
```diff
@@ -1,75 +1,149 @@
...
+
+    questionnaire: Questionnaire
+    responses: list[tuple[str, list]]
+
+    @validator("responses", each_item=True)
+    def validate_responses(
+        response: tuple[str, list], values: dict[str, Questionnaire]
+    ):
+        questionnaire = values.get("questionnaire")
+        question_name = response[0]
```
jaklinger, line 103: rather than passing the response tuple into both functions better to unpack here

```python
question_name, answers = response
```

and then down below do

```
validate_response_against_question(question=question, answers=answers)
```

src/layers/domain/core/questionnaire.py
```diff
@@ -1,75 +1,149 @@
...
+            f"Question '{question.name}' does not allow multiple responses. Response given: {answers}."
+        )
+
+    errors = (
+        []
+    )  # accumulate errors here for multianswer question and raise under a single ValueError
+    for answer in answers:
+        if not isinstance(answer, question.type):
+            errors.append(
+                f"Question '{question.name}' expects type {question.type}. Response given: {answer}."
```
jaklinger, line 129: just to help debugging:

```suggestion
                f"Question '{question.name}' expects type {question.type}. Response '{answer}' is of type '{type(answer)}'."
```

src/layers/domain/core/questionnaire.py
```diff
@@ -1,75 +1,149 @@
...
+            )
+
+        if question.validation_rules is not None:
+            for validation_rule in question.validation_rules:
+                try:
+                    validation_rule(answer)
+                except ValueError as e:
+                    errors.append(
+                        f"Question '{question.name}' validation failed for response: {answer}. Error: {str(e)}"
```
jaklinger, line 138: Adding a name of the rule for debugging purposes

```suggestion
                        f"Question '{question.name}' rule '{validation_rule.__name__}' failed validation for '{answer}' with error: {e}"
```

src/layers/domain/core/questionnaire.py
```diff
@@ -1,75 +1,149 @@
...
-    """
-    The data type of the question
-    """
-
-    STRING = str
-    INT = int
-    BOOL = bool
-    DATE_TIME = datetime
+T = TypeVar("T")
+A = {str, int, bool, datetime, float, date, time}  # allowed question types
```
jaklinger, line 12: tiny one - but I'd renamed this to `ALLOWED_QUESTION_TYPES` and then don't need the comment

jaklinger (approved):

PR #134, opened 2024-02-06
[Feature PI-45-dns] Added in namespace delegation and PE zones

Rohoolio:


infrastructure/terraform/per_account/mgmt/terraform.tf
```diff
@@ -0,0 +1,18 @@
+terraform {
+  backend "s3" {
+    encrypt              = false
+    region               = "eu-west-2"
+    bucket               = "nhse-cpm--terraform-state-v1.0.0"
+    dynamodb_table       = "nhse-cpm--terraform-state-lock-v1.0.0"
+    key                  = "terraform-state-infrastructure-mgmt-dns"
```
jaklinger, line 7: is dns the right suffix here?

jaklinger (approved):
some minor comments, otherwise lgtm

infrastructure/terraform/per_account/mgmt/terraform.tf
Rohoolio: As discussed this is fine for now as it will cause issues with the bucket redeploying all our zones

PR #146, opened 2024-02-23
Feature/pi 192 spine device and endpoint questionnaires

megan-bower4:


src/layers/domain/core/spine_device_questionnaire.py
jaklinger, line 1: I think we should rethink this a little - better to store data in data files, so I'd suggest that we switch to something like

`questionnaires/spine_device/v1.json`:

```json
[
    {"name": "nhs_mhs_manufacturer_org", ..., "answer_type": "str"}
...
]
```

and we have a small tool for converting questionnaire json into `Questionnaire` instances:

```python
def render_questionnaire(questionnaire_name, questionnaire_version, questions: list[dict]) -> Questionnaire:
    questionnaire = Questionnaire(name=questionnaire_name, version=questionnaire_version)
    for question in questions:
        questionnaire.add_question(**question)
    return questionnaire
```

where `questionnaire_name` and `questionnaire_version` comes from the path to the json

megan-bower4: Have added this functionality in the latest commit after working through it with Joel

src/layers/domain/core/questionnaire.py
```diff
@@ -24,14 +24,24 @@ class Config:
...
+            invalid_types = [
+                item for item in answer_type if item not in ALLOWED_QUESTION_TYPES
+            ]
+            if invalid_types:
+                raise ValueError(
+                    f"Answer types {', '.join(map(str, invalid_types))} are not allowed."
+                )
+        elif answer_type not in ALLOWED_QUESTION_TYPES:
             raise ValueError(f"Answer type {answer_type} is not allowed.")
         return answer_type
```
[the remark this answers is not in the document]
megan-bower4: answer_types now also defaults to a set to simplify this even more

src/layers/domain/core/load_questionnaire.py
```diff
@@ -0,0 +1,53 @@
...
+from event.json import json_load
+
+from .questionnaire import ALLOWED_QUESTION_TYPES, Questionnaire
+
+PATH_TO_HERE = Path(__file__).parent
+
+VALIDATION_RULE_MAPPING = {
+    "empty_str": empty_str,
+    "url": url,
+}
```
jaklinger, line 13: sorry realised that we can avoid having to add new rules manually with:

```suggestion
from domain.core import questionnaire_validation_custom_rules
from event.json import json_load

from .questionnaire import ALLOWED_QUESTION_TYPES, Questionnaire

PATH_TO_HERE = Path(__file__).parent
```

and then below do

```python
def process_validation_rules(validation_rules):
    return [questionnaire_validation_custom_rules.__dict__[rule] for rule in validation_rules]
```

src/layers/domain/core/load_questionnaire.py
```diff
@@ -0,0 +1,53 @@
...
+    json_file_path = f"{PATH_TO_HERE}/questionnaires/{questionnaire_name}/v{questionnaire_version}.json"
+    questions = load_json_file(json_file_path)
+    questionnaire = Questionnaire(
+        name=questionnaire_name, version=questionnaire_version
+    )
+
+    for question in questions:
+        # Convert answer_type string to Python type
+        if "answer_types" in question:
+            question["answer_types"] = process_answer_types(question["answer_types"])
```
jaklinger, line 44: can remove the comment if we name the function `convert_type_names_to_python_types`

jaklinger: similarly could name the below function to `convert_rule_names_to_rule_functions`

jaklinger (approved):
lgtm, couple of minor suggestions

PR #148, opened 2024-02-29
[PI-240] dns_to_domain 

Rohoolio:
Added domain code for dns to api gateway

### DIFF
--- a/infrastructure/terraform/per_account/mgmt/modules/route53/main.tf
+++ b/infrastructure/terraform/per_account/mgmt/modules/route53/main.tf
@@ -37,27 +37,27 @@ module "zones" {
 
 }
 
-resource "aws_route53_record" "prod_zone" {
-  zone_id = module.zones.route53_zone_zone_id["cpm.national.nhs.uk"]
-  name    = "api.cpm.national.nhs.uk"
-  records = [
-    "ns-453.awsdns-56.com.",
-    "ns-980.awsdns-58.net.",
-    "ns-1983.awsdns-55.co.uk.",
-    "ns-1103.awsdns-09.org."
-  ]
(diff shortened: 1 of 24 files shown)


infrastructure/terraform/per_account/mgmt/modules/route53/main.tf
```diff
@@ -37,27 +37,27 @@ module "zones" {
...
+#   name    = "api.cpm.national.nhs.uk"
+#   records = [
+#     "ns-453.awsdns-56.com.",
+#     "ns-980.awsdns-58.net.",
+#     "ns-1983.awsdns-55.co.uk.",
+#     "ns-1103.awsdns-09.org."
+#   ]
+#   ttl  = 300
+#   type = "NS"
+# }
```
megan-bower4, line 51: How come we don't need a prod zone? And do we need the commented out code still? Or will we be adding it back at a later date so makes sense to keep the work done already?

infrastructure/terraform/per_workspace/modules/api_entrypoint/vars.tf
```diff
@@ -15,3 +15,6 @@ variable "project" {
 
 variable "authoriser_metadata" {
 }
+
+variable "domain" {
+}
```
megan-bower4, line 20: Very small one - sometimes type is specified and sometimes it's not? Is this just a preference?

Rohoolio: I think this is probably an inconsistency thing on our part, most of the ones I look at that do have a type have differed from that of a string so we should probably align everything soon to at least include a type haha

PR #166, opened 2024-03-20
[feature/PI-213-connect_to_ldap]

megan-bower4:


infrastructure/terraform/per_account/dev/main.tf
```diff
@@ -45,6 +45,24 @@ module "bucket" {
   }
 }
 
+module "truststore_bucket" {
```
Rohoolio, line 48: We already have a truststore in mgmt thats creating during the bootstrap scripts to use so id say use that, unless you plan to have one in every environment which I think would be overkill for storing certs

Rohoolio: Ah ok looking back at it now - it seems that we store everything in the mgmt ones and then pull them down as part of the build to then push them out to env specific buckets - ignore me :D

Rohoolio (approved):

PR #182, opened 2024-04-25
[Feature/PI-268] Search for a device

jameslinnell:


src/api/searchDevice/src/v1/steps.py
```diff
@@ -1,15 +1,80 @@
...
+from domain.fhir_translation.device import create_fhir_searchset_bundle
+from domain.repository.device_repository import DeviceRepository
+from domain.repository.marshall import unmarshall_value
 from event.step_chain import StepChain
 
-from ..data.response import devices, endpoints
-
 
-def get_results(data, cache):
+# Think we require a validation decorator here.
```
Rohoolio, line 11: Something we should deal with now or are we happy to continue without this?

src/api/searchDevice/src/v1/steps.py
```diff
@@ -1,15 +1,80 @@
...
+        if event.query_string_parameters["device_type"]
+        else None
+    )
+
+
+def set_device_type(data, cache):
+    event_data = data[parse_event_query]
+    device_query_param = event_data.get("query_string")
+    return (
+        DeviceType.PRODUCT if device_query_param == "PRODUCT" else DeviceType.ENDPOINT
```
Rohoolio, line 28: Would it be better to write this so that its not one or the other - we are very likely to have different devicetypes in the future - could do a map of string to devicetype and get the right devicetype from that?

src/api/searchDevice/src/v1/steps.py
jameslinnell: Yes, should do that

src/api/tests/feature_tests/steps/requests.py
```diff
@@ -103,7 +103,7 @@ def _mocked_request(
         path=url,
         httpMethod=method,
         headers=headers,
-        multiValueHeaders={},
+        multiValueHeaders={"Host": ["foo.co.uk"]},
```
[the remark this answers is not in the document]
Rohoolio: ah nevermind its for the fhir translation 👍

src/api/searchDevice/src/v1/steps.py
jameslinnell: It's not the case now. I guess until that happens we should leave it. That's agile.

src/api/tests/feature_tests/steps/requests.py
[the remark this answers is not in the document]
jameslinnell: For building urls in the response

src/api/searchDevice/src/v1/steps.py
jameslinnell: In theory it won't get this far as there is now validation on the device_type filter. Anything other than "product" or endpoint will get rejected. However I've now set it up to be a map. which can be added to.

Rohoolio (approved):

PR #189, opened 2024-05-01
[Feature/PI-321] Search only returns active devices

jameslinnell:


Rohoolio (approved):

src/layers/domain/response/validation_errors.py
```diff
@@ -70,14 +77,60 @@ def decorator(*args: P.args, **kwargs: P.kwargs):
...
+
+    @classmethod
+    def _missing_(cls, value):
+        for member in cls:
+            if value.upper() == member.value:
+                return member
+
+
+class SearchQueryParams(BaseModel, extra=Extra.forbid):
+    device_type: DeviceType
```
jaklinger, line 92: 1. I think it would be good to import DeviceType from src/layers/domain/core/device.py instead
2. Can move the case-insensitivity to

```
class SearchQueryParams(BaseModel, extra=Extra.forbid):
    device_type: DeviceType

    @validator("device_type")
    def validate_device_type(device_type: str):
        return device_type.lower()

3. Would move these to a new module `layers/domain/api/query.py`

File: src/layers/domain/response/validation_errors.py [hunk 2]
```diff
@@ -70,14 +77,60 @@ def decorator(*args: P.args, **kwargs: P.kwargs):
...
+        try:
+            SearchQueryParams(**query_params)
+            return function(data, cache, **kwargs)
+        except ValidationError as exc:
+            raise InboundQueryValidationError(
+                errors=exc.raw_errors,
+                model=exc.model,
+            )
+
+    return decorator
```
jaklinger, line 109: I'd write this code directly into the search step as it's not really a part of the response validation errors utils. 

Also don't think we need a specific `InboundQueryValidationError` - I can see below that you've set the message to `"'device_type' query parameter must be one of 'product' or 'endpoint'` but that's already pretty close to the default message

```
value is not a valid enumeration member; permitted: 'product', 'endpoint' (type=type_error.enum; enum_values=[<DeviceType.PRODUCT: 'product'>, <DeviceType.ENDPOINT: 'endpoint'>])
```

jaklinger (approved):

PR #191, opened 2024-05-01
[PI-335] changelog delete

jaklinger:


src/layers/sds/cpm_translation/__init__.py
```diff
@@ -71,24 +56,34 @@ def translate(
...
             repository=repository,
         )
-        translations = MESSAGE_HANDLING_SYSTEM_TRANSLATIONS
+    elif object_class == SdsDeletionRequest.OBJECT_CLASS:
+        instance = SdsDeletionRequest.construct(**obj)
+        devices = delete_devices(
+            deletion_request=instance,
+            questionnaire_ids=[
+                spine_endpoint_questionnaire.id,
+                spine_device_questionnaire,
```
Rohoolio, line 80: think this is missing the id?

src/layers/domain/core/device.py
```diff
@@ -132,6 +170,27 @@ def delete_key(self, key: str) -> DeviceKeyDeletedEvent:
...
+        self, questionnaire_id: str, question_name: str
+    ) -> list[DeviceIndexAddedEvent]:
+        questionnaire_responses = _get_questionnaire_responses(
+            questionnaire_responses=self.questionnaire_responses,
+            questionnaire_id=questionnaire_id,
+        )
+        unique_answers = _get_unique_answers(
+            questionnaire_responses=questionnaire_responses, question_name=question_name
+        )
+        index_added_event = partial(
```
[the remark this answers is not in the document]
jaklinger: background:
* for a device, for a specific questionnaire, there may be multiple questionnaire responses

src/layers/sds/cpm_translation/__init__.py
jaklinger: thanks! not actually run/written any tests or anything yet...

src/etl/sds/tests/test_sds_etl_components.py
```diff
@@ -24,6 +24,21 @@
 from test_helpers.pytest_skips import long_running
 from test_helpers.terraform import read_terraform_output
 
+# Note that unique identifier "000428682512" is the identifier of 'GOOD_SDS_RECORD'
+DELETION_REQUEST_000428682512 = """
+dn: o=nhs,ou=Services,uniqueIdentifier=000428682512
+changetype: delete
+objectclass: delete
```
Rohoolio, line 31: Im just working my way down the PR so i may get my answer further down - are we making the objectclass delete to trigger something further down? If i recall these are normally the top, changelogentry etc thats the same for all the changelog stuff from ldap

src/etl/sds/tests/test_sds_etl_components.py
```diff
@@ -206,3 +227,88 @@ def test_end_to_end_bulk_trigger(repository: MockDeviceRepository):
...
+    )
+    assert device.status == DeviceStatus.INACTIVE
+
+    # Verify that the other device is now "inactive"
+    (device,) = repository.read_by_index(
+        questionnaire_id="spine_device/1",
+        question_name="unique_identifier",
+        value="000842065542",
+    )
+    assert device.status == DeviceStatus.INACTIVE
```
Rohoolio, line 314: I love this test, so easy to follow and super descriptive <3

src/layers/sds/cpm_translation/__init__.py
```diff
@@ -67,28 +43,38 @@ def translate(
...
+        nhs_mhs = NhsMhs.construct(**obj)
+        devices = [
+            nhs_mhs_to_cpm_device(
+                nhs_mhs=nhs_mhs,
+                questionnaire=spine_endpoint_questionnaire,
+                _questionnaire=_spine_endpoint_questionnaire,
+                _trust=_trust,
+            )
+        ]
+    elif object_class == SdsDeletionRequest.OBJECT_CLASS:
```
Rohoolio, line 65: found my answer to the first question I made 😂

src/layers/sds/cpm_translation/tests/test_cpm_translation.py
```diff
@@ -60,7 +74,20 @@
 )
 
 
-@settings(deadline=1000)
+@pytest.fixture
+def repository(
+    request: pytest.FixtureRequest,
+) -> Generator[DeviceRepository, None, None]:
+    if request.node.get_closest_marker("integration"):
```
Rohoolio, line 81: get_closest_marker always makes me feel a bit nervous, an awkward name really

src/layers/sds/domain/changelog.py
```diff
@@ -50,13 +50,22 @@ def validate_change_number(
...
+        header_lines = [
             f"dn: {self.target_distinguished_name.raw}",
             f"changetype: {self.change_type}",
         ]
-        if self.change_type is not ChangeType.DELETE:
-            change_lines.append(self.changes.strip("\n"))
+        if self.change_type is not ChangeType.ADD:
+            header_lines.append(f"{OBJECT_CLASS_FIELD_NAME}: {self.change_type}")
+
+        change_lines = header_lines + list(filter(bool, self.changes.split("\n")))
```
Rohoolio, line 67: Will this fail if changes is missing? I found one or two cases where instead of the changes object there looked to be an objectClass name (its int data so could be bogus but it happened at least once so I made changes optional above)

src/etl/sds/tests/test_sds_etl_components.py
jaklinger: ah yeah the hook that was already implemented in the `extract` worker is to filter/hook ldif objects by `objectclass` (i.e. `objectClass: nhsMhs` is mapped on to our python model `NhsMhs`)

This way we can create a new object class `delete` and filter/map it on to the new `SdsDeletionRequest` object

jaklinger: and using it as the same value as `changetype` is just convenient, as the same principle can apply to `modify`

src/layers/sds/domain/changelog.py
jaklinger: the failure would happen in the parsing into a model, so for `SdsDeletionRequest` no changes are expected, `SdsModificationRequest` at least one field is expected and regular `add` expressions all the required fields are expected

but if any of those fail then it means that the ldif is bad, so we'd want it to fail at the model parsing at least

Rohoolio (approved):

PR #195, opened 2024-05-15
[Feature/PI-320] Device created on and deleted on date

jameslinnell:


src/layers/domain/core/device.py
```diff
@@ -46,6 +51,9 @@ class DeviceUpdatedEvent(Event):
     product_team_id: UUID
     ods_code: str
     status: "DeviceStatus"
+    created_on: str
+    updated_on: Optional[str] = None
```
Rohoolio, line 55: Isnt a device updated event always going to have an updated_on date?

jameslinnell: Yes, good spot

Rohoolio (approved):

PR #198, opened 2024-05-17
Feature/pi 207 state lock etl

megan-bower4:


src/etl/sds/executor/input/tests/test_input_trigger.py
```diff
@@ -0,0 +1,575 @@
...
+from etl_utils.trigger.model import _create_timestamp
+from etl_utils.trigger.operations import StateFileNotEmpty
+from moto import mock_aws
+from mypy_boto3_s3 import S3Client
+from mypy_boto3_stepfunctions import SFNClient
+
+from etl.sds.executor.input.steps import _start_execution
+from test_helpers.terraform import read_terraform_output
+
+MOCKED_INPUT_TRIGGER_ENVIRONMENT = {
```
Rohoolio, line 25: Might be worth moving all these constants to a separate file so that the tests are the main star of this file?

infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
```diff
@@ -432,3 +440,36 @@ module "bulk_trigger_notification" {
...
+    # prepend '/opt/python' (the root path of the layer) to LD_LIBRARY_PATH so that
+    # all compiled dependencies can find each other. Note: this is a hack - and
+    # may result in version mismatches between system libs on the lambda. The stable
+    # alternative is to run or deploy the service from a container.
+    LD_LIBRARY_PATH         = "/opt/python:/var/lang/lib:/lib64:/usr/lib64:/var/runtime:/var/runtime/lib:/var/task:/var/task/lib:/opt/lib"
+    TRUSTSTORE_BUCKET       = var.truststore_bucket.id
+    CPM_FQDN                = "cpm.thirdparty.nhs.uk"
+    LDAP_HOST               = data.aws_secretsmanager_secret_version.ldap_host.secret_string
+    LDAP_CHANGELOG_USER     = data.aws_secretsmanager_secret_version.ldap_changelog_user.secret_string
+    LDAP_CHANGELOG_PASSWORD = data.aws_secretsmanager_secret_version.ldap_changelog_password.secret_string
```
jaklinger, line 470: i think most of these env vars aren't needed for the state lock lambda?

infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
```diff
@@ -432,3 +440,36 @@ module "bulk_trigger_notification" {
   filter_prefix = "${local.bulk_trigger_prefix}/"
   filter_suffix = ".ldif"
 }
+
+module "proxy_lambda_executor" {
+  source = "./proxy-executor/"
```
jaklinger, line 445: i'm wondering whether we should give a name that helps the developer pin down where this fits into the ETL. maybe "ETL state lock enforcer" so something like `module etl_state_lock_enforcer` and `source = ./etl-state-lock-enforcer`

src/etl/sds/executor/input/input.py
```diff
@@ -0,0 +1,64 @@
...
+
+class ChangelogTriggerEnvironment(BaseEnvironment):
+    STATE_MACHINE_ARN: str
+    NOTIFY_LAMBDA_ARN: str
+    TRUSTSTORE_BUCKET: str
+    CPM_FQDN: str
+    LDAP_HOST: str
+    ETL_BUCKET: str
+    LDAP_CHANGELOG_USER: str
+    LDAP_CHANGELOG_PASSWORD: str
```
jaklinger, line 18: I think we can ditch a lot of these environment vars?

src/etl/sds/executor/input/input.py
```diff
@@ -0,0 +1,64 @@
+import boto3
+from etl_utils.trigger.logger import log_action
+from etl_utils.trigger.notify import notify
+from event.environment import BaseEnvironment
+from event.step_chain import StepChain
+
+from .steps import _process_sqs_message, steps
+
+
+class ChangelogTriggerEnvironment(BaseEnvironment):
```
jaklinger, line 10: `StateLockEnvironment` ?

infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
```diff
@@ -432,3 +440,36 @@ module "bulk_trigger_notification" {
...
+  workspace_prefix      = var.workspace_prefix
+  python_version        = var.python_version
+  event_layer_arn       = var.event_layer_arn
+  third_party_layer_arn = var.third_party_sds_update_layer_arn
+  sds_layer_arn         = var.sds_layer_arn
+  etl_bucket_arn        = module.bucket.s3_bucket_arn
+  etl_layer_arn         = module.etl_layer.lambda_layer_arn
+  notify_lambda_arn     = module.notify.arn
+  state_machine_arn     = aws_sfn_state_machine.state_machine.arn
+  table_arn             = var.table_arn
```
jaklinger, line 458: don't think we need table arn?

infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
```diff
@@ -432,3 +440,36 @@ module "bulk_trigger_notification" {
   filter_prefix = "${local.bulk_trigger_prefix}/"
   filter_suffix = ".ldif"
 }
+
+module "proxy_lambda_executor" {
+  source = "./proxy-executor/"
+
+  executor_name         = "input"
```
jaklinger, line 447: maybe keep the same naming convention all the way down, so that the lambda has the same name as the module. Would help to draw the developers eye to parts of the codebase that are related

jaklinger: imo not immediately clear that this needs to be a parameter. I think unless the parameters can't be hard-coded into the module probably best to hard code them into the module as by definitions this isn't a re-use module (i.e. we only want one state locking mechanism for this etl). Might be an idea to have variables like this as `locals` instead inside the module, especially if they're repeatedly used in the module

src/etl/sds/executor/input/input.py
jaklinger, line 10: Remove word `Changelog`

src/etl/sds/executor/input/input.py
```diff
@@ -0,0 +1,64 @@
...
+    "state_machine_arn": ENVIRONMENT.STATE_MACHINE_ARN,
+    "etl_bucket": ENVIRONMENT.ETL_BUCKET,
+}
+
+
+def handler(event=dict, context=None):
+    for message in event["Records"]:
+        process_message(message)
+
+    return None  # Message removed from sqs queue
```
jaklinger, line 39: I'd add a more verbose comment here to warn not to change this return value as it's the expected return from a successful SQS lambda

src/etl/sds/trigger/bulk/steps.py
```diff
@@ -42,49 +39,56 @@ def _validate_database_is_empty(data, cache: Cache):
...
 
-def _delete_object(data, cache: Cache):
+def _delete_trigger_object(data, cache: Cache):
     source_bucket, source_key = data[StepChain.INIT]
     return cache["s3_client"].delete_object(Bucket=source_bucket, Key=source_key)
 
 
-def _start_execution(data, cache):
+def _publish_message_to_sqs_queue(data, cache: Cache):
+    sqs_client = boto3.client("sqs")
```
jaklinger, line 62: best practice is to declare boto3 clients at the handler level of the lamdba and cache them (e.g. with `s3_client`), although in this case (because bulk is only run once) this is defo v pedantic!

src/etl/sds/trigger/bulk/steps.py
```diff
@@ -42,49 +39,56 @@ def _validate_database_is_empty(data, cache: Cache):
...
-        state_machine_input=state_machine_input,
+    state_machine_name = state_machine_input.name
+    message_body = {
+        "changelog_number_start": state_machine_input.changelog_number_start,
+        "changelog_number_end": state_machine_input.changelog_number_end,
+        "etl_type": state_machine_input.etl_type.value,
+        "timestamp": state_machine_input.timestamp,
+        "name": state_machine_name,
+    }
+    message_body_json = json.dumps(message_body)
```
jaklinger, line 74: is this any different to

```python
state_machine_input.json()
```
?

src/etl/sds/trigger/update/steps.py
```diff
@@ -113,32 +109,39 @@ def _parse_and_join_changelog_changes(data, cache):
...
-    latest_changelog_number = data[_get_latest_changelog_number_from_ldap]
-    changes = data[_parse_and_join_changelog_changes]
+    # Convert the state_machine_input to JSON
     state_machine_input: StateMachineInput = data[_create_state_machine_input]
-
-    return cache["s3_client"].put_object(
-        Bucket=cache["etl_bucket"],
-        Key=f"history/{state_machine_input.etl_type}/{latest_changelog_number}/{WorkerKey.EXTRACT}",
-        Body=changes,
+    message_body = {
```
jaklinger, line 128: also just to check whether could do `state_machine_input.json()`

infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
megan-bower4: Ah yes, lots are unnecessary. Also removed the step function info/permissions from the bulk/update triggers as only the input lambda needs this now.

src/etl/sds/trigger/bulk/steps.py
megan-bower4: Ah yeah that is much nicer! It just doesn't include the name parameter so I've added a little function to the StateMachine class to return the json including the name as well.

src/etl/sds/executor/input/steps.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class Cache(TypedDict):
+    s3_client: "S3Client"
+    step_functions_client: "SFNClient"
+    state_machine_arn: str
+    ldap: LdapModuleProtocol
+    ldap_host: str
+    cert_file: Path
+    key_file: Path
```
jaklinger, line 31: these aren't in this lambda?

jaklinger (approved):

PR #201, opened 2024-05-20
[PI-346a] LDIF modify parsing

jaklinger:


jameslinnell (approved):

src/layers/etl_utils/ldif/_ldif.py
```diff
@@ -28,7 +30,9 @@
 safe_string_re = re.compile(SAFE_STRING_PATTERN)
 
 
-CHANGE_TYPES = ["add", "delete", "modify", "modrdn"]
+CHANGE_TYPES = {"add", "delete", "modify"}  # ,'modrdn'
```
megan-bower4, line 33: Can we remove modrdn?

jaklinger: given that the `_ldif` code is pulled from an open source project and we don't know loads about ldif ourselves, i though maybe to keep `modrdn` in comment in case it turns out to be something we need. I've made an assumption that `modrdn` never comes up as it would be very weird to change the `dn` in SDS, but just in case I thought it would be useful to show that we're actively not considering it at present

PR #204, opened 2024-05-21
[PI-346d] modify transform and load

jaklinger:


src/layers/sds/cpm_translation/translations.py
```diff
@@ -156,10 +122,65 @@ def nhs_mhs_to_cpm_device(
...
+        field_name = model.get_field_name_for_alias(alias=field)
+
+        try:
+            modify_key = get_modify_key_function(
+                model=model,
+                field_name=field_name,
+                modification_type=modification_type,
+            )
+            _devices += list(
+                modify_key(devices=_devices, field_name=field_name, value=new_values)
```
jaklinger, line 170: check that this doesn't lead to duplicate events

jameslinnell (approved):

src/layers/sds/domain/base.py
```diff
@@ -67,6 +73,50 @@ def _validate_object_class(cls: "SdsBaseModel", values: dict) -> dict:
...
+            field_name
+            for field_name, model_field in cls.__fields__.items()
+            if model_field.alias == alias
+        )
+    except ValueError:
+        raise AliasLookupError(f"No field with alias '{alias}' found")
+    return field_name
+
+
+def _get_alias_for_field_name(cls: BaseModel, field):
```
megan-bower4, line 88: Just a question for my understanding - what's the purpose of the alias'? And like what scenario would they be used in?

jaklinger: in the models, the "aliases" are the names coming from the ldap server (e.g. `nhsasclient`) whereas the field names are a bit more readable (`nhs_as_client`). When doing `add` operations we pass the whole lot in for pydantic to sort out (so `nhsasclient` gets converted to `nhs_as_client`) but for modify we're given `nhsasclient` and then we need to manually (via the model) to convert to `nhs_as_client`

PR #213, opened 2024-05-29
[Feature PI-342] redact logging fields

Rohoolio:


src/layers/event/logging/logger.py
```diff
@@ -1,15 +1,17 @@
 from pathlib import Path
+from typing import Set
 from uuid import uuid4
 
 from nhs_context_logging import app_logger
 
 
-def setup_logger(service_name: str, uuid: str = None):
+def setup_logger(service_name: str, uuid: str = None, redacted_fields: Set[str] = None):
```
jaklinger, line 8: doesn't matter, but these days you can just use `set` rather than the imported `Set`

jaklinger (approved):

PR #215, opened 2024-05-30
[Feature PI-318] cache_zips_and_venv 

Rohoolio:
Cache for third party build

### DIFF
--- a/.github/actions/pytest-with-rerun/action.yml
+++ b/.github/actions/pytest-with-rerun/action.yml
@@ -15,9 +15,12 @@ runs:
       with:
         command: test--${{ inputs.test-type }}
         requires-aws: ${{ inputs.requires-aws }}
+        restore-from-cache: "false"
+        restore-third-party-cache: "true"
 
     - uses: ./.github/actions/make/
       with:
         command: test--${{ inputs.test-type }}--rerun PYTEST_FLAGS="--suppress-no-test-exit-code"
         requires-aws: ${{ inputs.requires-aws }}
         restore-from-cache: "false"
+        restore-third-party-cache: "true"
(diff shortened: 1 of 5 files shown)


.github/actions/pytest-with-rerun/action.yml
```diff
@@ -15,9 +15,12 @@ runs:
       with:
         command: test--${{ inputs.test-type }}
         requires-aws: ${{ inputs.requires-aws }}
+        restore-from-cache: "false"
+        restore-third-party-cache: "true"
```
jaklinger, line 19: i would've thought we'd just want the regular cache here

.github/actions/pytest-with-rerun/action.yml
```diff
@@ -15,9 +15,12 @@ runs:
...
 
     - uses: ./.github/actions/make/
       with:
         command: test--${{ inputs.test-type }}--rerun PYTEST_FLAGS="--suppress-no-test-exit-code"
         requires-aws: ${{ inputs.requires-aws }}
         restore-from-cache: "false"
+        restore-third-party-cache: "true"
```
jaklinger, line 26: don't think we want this here

.github/workflows/_deploy.yml
```diff
@@ -84,6 +85,7 @@ jobs:
           workspace: ${{ env.WORKSPACE }}
           scope: ${{ env.SCOPE }}
           restore-from-cache: "true"
+          restore-third-party-cache: "true"
```
jaklinger, line 88: don't think we need this anywhere except in build?

.github/actions/pytest-with-rerun/action.yml
Rohoolio: If you take it away it cant run pytest or behave as it fails to locate the package - likely because the .venv is part of the cache

jaklinger (approved):

PR #260, opened 2024-07-11
[feature/PI-457-update_loop] wrap 10x loop around update sub-etl

jaklinger:


infrastructure/terraform/per_workspace/modules/etl/sds/etl-diagram--update-transform-and-load.asl.json
```diff
@@ -1,91 +1,113 @@
...
-      "Next": "etl-stopped"
+      }
     },
-
-    "etl-stopped": {
+    "pass": {
       "Type": "Pass",
-      "End": true
+      "End": true,
+      "OutputPath": "$[9]"
```
megan-bower4, line 110: What happens if it does't need to loop 10 times? Or will it always loop the full amount and split the changes/events evenly?

jaklinger: it will always loop the full amount, and if any are not needed then they effectively won't do anything

megan-bower4 (approved):

PR #265, opened 2024-07-16
[feature/PI-444-domain_v2] add domain v2

jaklinger:


Rohoolio (approved):

src/layers/domain/core/device/v2.py
```diff
@@ -0,0 +1,312 @@
...
+
+class QuestionNotFoundError(Exception):
+    pass
+
+
+@dataclass(kw_only=True, slots=True)
+class DeviceCreatedEvent(Event):
+    id: str
+    name: str
+    type: "DeviceType"
```
megan-bower4, line 41: device_type

src/layers/domain/core/device_key/tests/test_device_key_v1.py
megan-bower4, line 1: Do we need a test_device_key_v2.py as well?

src/layers/domain/core/device_key/v2.py
```diff
@@ -0,0 +1,17 @@
...
+    A Device Key is a secondary way of indexing / retrieving Devices
+    """
+
+    key_type: DeviceKeyType
+    key_value: str
+
+    @validator("key_value", check_fields=True)
+    def validate_key(key_value: str, values: dict):
+        key_type: DeviceKeyType = values.get("key_type")
+        return validate_key(key=key_value, type=key_type)
```
megan-bower4, line 17: Refactor the validate_key function outside of this too?


```python
def validate_key(key_value: str, key_type: DeviceKeyType):
    if key_type and key_type.pattern.match(key) is None:
        raise InvalidDeviceKeyError(
            f"Key '{key_value}' does not match the expected "
            f"pattern '{key_type.pattern.pattern}' associated with "
            f"key type '{key_type}'"
        )
    return key_value
```

src/layers/domain/core/device_key/tests/test_device_key_v1.py
jaklinger: i think it's ok because everything tested in `test_device_key_v1.py` is directly imported in v2, so we wouldn't be testing anything new. Important consideration when removing v1 would be to migrate the tests, but i'd think that would be clear given that the functional code would need to be ported as well!

src/layers/domain/core/device_key/v2.py
jaklinger: yeah wasn't totally sure which side of the fence i sat on this one. Reason to not the private-ish internal function is so that the behaviour between versions is coupled without duplicated code. Feels like it would be easy to introduce bugs between versions if we duplicated too much

PR #266, opened 2024-07-17
[feature/PI-447] Search SDS Endpoints and Devices

jameslinnell:


src/layers/domain/repository/device_repository/v2.py
```diff
@@ -0,0 +1,45 @@
...
+    device_5NR_result,
+    device_RTX_result,
+    endpoint_5NR_result,
+    endpoint_RTX_result,
+    no_device_results,
+    no_endpoint_results,
+)
+
+
+class DeviceRepository(Repository[Device]):
```
Rohoolio, line 14: This looks like it was fun to do haha

PR #271, opened 2024-07-25
[PI-446] repository v2

megan-bower4:


src/layers/domain/core/device_key/v2.py
```diff
@@ -14,4 +15,21 @@ class DeviceKey(BaseModel):
...
+
+    def __hash__(self):
+        return hash(self.parts)
+
+
+def validate_key(key_value: str, key_type: DeviceKeyType):
+    if key_type and key_type.pattern.match(key_value) is None:
+        raise InvalidDeviceKeyError(
+            f"Key '{key_value}' does not match the expected "
+            f"pattern '{key_type.pattern.pattern}' associated with "
```
Rohoolio, line 32: Not even a real issue but pattern.pattern looks a bit funny

src/layers/domain/repository/device_repository/v2.py
```diff
@@ -0,0 +1,190 @@
...
+    ]
+    return [root_primary_key] + non_root_primary_keys
+
+
+def handle_update_device(
+    table_name: str, id: str, keys: list[DeviceKey], data: dict
+) -> list[TransactItem]:
+    primary_keys = _device_primary_keys(device_id=id, device_keys=keys)
+    update_expression = _dynamodb_update_expression(updated_fields=data)
+    update_statement = partial(
```
Rohoolio, line 71: This is for my own understanding - how much value does having a partial here achieve? feels like if the update_statement was being used more than once then i can understand its use. I might be missing what this actually achieves though :D

src/layers/domain/repository/device_repository/v2.py
```diff
@@ -0,0 +1,190 @@
...
+    return TransactItem(
+        Put=TransactionStatement(
+            TableName=table_name,
+            Item=marshall(pk=key, sk=key, root=True, **device_data),
+            ConditionExpression=ConditionExpression.MUST_NOT_EXIST,
+        )
+    )
+
+
+def handle_delete_device(table_name: str, key_parts: tuple[str]) -> TransactItem:
```
Rohoolio, line 96: I dont think we had a delete device mechanism before so I suppose this comment doesnt matter too much, but i dont think this will this go and delete all the extra rows that belong to that "device" will it?

src/layers/domain/repository/repository/v2.py
Rohoolio, line 1: thank god for Chat GPT which helped my smooth brain understand this file haha

src/layers/domain/core/device_key/v2.py
jaklinger: agreed, but it's an object of python type `re.Pattern` and it has a member called `pattern`

src/layers/domain/core/root/tests/test_root_v2.py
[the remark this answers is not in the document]
jaklinger: oh that should have a test in it 😳

src/layers/domain/repository/device_repository/v2.py
jaklinger: it deletes a given single device row by key - so the intention is to delete either a tag-indexed or a key-indexed device - not the root device. If we ever accidentally used this function actually hard delete root devices that would be caught by integration tests, as there all tests currently indirectly check that no device is hard deleted anyway

src/layers/domain/repository/repository/v2.py
jaklinger: 🧠

src/layers/domain/core/root/tests/test_root_v2.py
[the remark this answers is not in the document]
jaklinger: ok those are in, and removed a bug-like thing as a result 🎉

src/layers/domain/repository/device_repository/v2.py
Rohoolio: I think you wernt for device_index on the other PR which works!

PR #275, opened 2024-08-02
[feature/PI-464-repo_with_tags] repo with tags

jaklinger:


src/layers/domain/repository/device_repository/v2.py
```diff
@@ -81,20 +90,27 @@ def handle_update_device(
...
+    pk_key_parts: tuple[str],
+    device_data: dict,
+    sk_key_parts=None,
+    pk_table_key: TableKey = TableKey.DEVICE,
+    sk_table_key: TableKey = TableKey.DEVICE,
+    root=False,
 ) -> TransactItem:
-    key = TableKeys.DEVICE.key(*key_parts)
+    pk = pk_table_key.key(*pk_key_parts)
+    sk = sk_table_key.key(*sk_key_parts) if sk_key_parts else pk
```
Rohoolio, line 102: I think we need to adjust the name of this slightly cause its not creating a device each time its being called, we only have one "Device" and a bunch of copies we make for our own convenience. I wonder if handle_create_search_row or something along those lines would be better. Might also help avoid the issue where someone uses this to make an unusual device that isnt the CPM UUID

jaklinger: i'd go for `handle_new_device_index`

PR #276, opened 2024-08-02
[PI-285] repo_questionnaire_v2

jaklinger:


src/layers/domain/core/device/v2.py
```diff
@@ -266,7 +249,7 @@ def add_tag(self, **kwargs):
     @event
     def add_questionnaire_response(
         self, questionnaire_response: QuestionnaireResponse
-    ) -> QuestionnaireResponseAddedEvent:
+    ) -> QuestionnaireResponseUpdatedEvent:
```
Rohoolio, line 252: Have we gone with UpdatedEvent over CreatedEvent because we will always have a default value?

Rohoolio (approved):

src/layers/domain/core/device/v2.py
jaklinger: because questionnaire responses no longer have their own row, technically we're just updating a pre-existing field on the device, and the handler for that will be the same regardless of the operation... so i thought the most generic description of that was an "update" since the convention for updates tends to be "update or create if not exists"

Rohoolio (approved):

PR #291, opened 2024-08-21
Feature/PI-476-split_etl

jaklinger:


infrastructure/terraform/per_workspace/modules/etl/sds/main.tf
```diff
@@ -188,6 +279,19 @@ module "worker_load" {
     EOT
 }
 
+module "worker_load_bulk_reduce" {
```
Rohoolio, line 282: This is a new third one?

PR #293, opened 2024-08-23
[feature/PI-245] Matrix testing

jameslinnell:


src/api/tests/sds_data_tests/test_sds_data.py
```diff
@@ -0,0 +1,259 @@
...
+
+@pytest.mark.parametrize(
+    "item",
+    _generate_test_data(
+        "sds_fhir_api.unique_queries.json",
+        test_count=int(os.getenv("TEST_COUNT", "10")),
+    ),
+)
+@pytest.mark.matrix
+def test_api_responses_match(item):
```
jaklinger, line 189: is it possible to split this test into two tests - one for prod, one non-prod? my brain's melting following the flow - for example the line with 

```cpm_status, cpm_body = _request_dev(item)```

why doesn't that need `use_cpm` setting?

I'd maybe also consider splitting the test into 200 and non-200s if that's possible but not sure if that's complicated with the input data!

jameslinnell: There is another test for non-200s already. But I still have to catch them here as the unique_queries file doesn't always produce 200s.

dev env var USE_CPM=1 therefore filter for use_cpm is not needed.

eventually we won't need any tests for accessing dev. I would suggest when that time comes we remove that condition.

jaklinger (dismissed):

PR #294, opened 2024-08-23
Feature/PI-475-load_in_parallel

jaklinger:


src/etl/sds/worker/transform_bulk/tests/test_transform_bulk_worker.py
```diff
@@ -134,38 +142,44 @@ def test_transform_worker_pass_no_dupes_max_records(
...
-    n_unprocessed_records = n_initial_unprocessed
-    while n_unprocessed_records > 0:
-        n_newly_processed_records_expected = min(n_unprocessed_records, max_records)
-        n_unprocessed_records_expected = (
-            n_unprocessed_records - n_newly_processed_records_expected
-        )
-        n_total_processed_records_expected += n_newly_processed_records_expected
+    expected_responses = []
+    expected_iterations = min(
+        ceil(n_initial_unprocessed / max_records), transform_bulk.FAN_OUT
```
Rohoolio, line 147: Is rounding this value up ever going to get it wrong?

src/layers/domain/repository/repository/v2.py
```diff
@@ -20,7 +24,44 @@
...
+
+def exponential_backoff_with_jitter(
+    n_retries, base_delay=0.1, min_delay=0.05, max_delay=5
+):
+    """Calculate the delay with exponential backoff and jitter."""
+    delay = min(base_delay * (2**n_retries), max_delay)
+    return random.uniform(min_delay, delay)
+
+
+def retry_with_jitter(max_retries=5, error=ClientError):
```
Rohoolio, line 43: Is this jitter a way to handle the issues you mentioned around the DB taking too much load and potentially failing?

src/etl/sds/worker/transform_bulk/tests/test_transform_bulk_worker.py
jaklinger: the rounding is deliberate, e.g. with `transform_bulk.FAN_OUT = 10` (which the hard-coded value) then the behaviour of that line is if there are 1007 initial records and each "fan out" processes `max_records=100` then  `ceil(1007 / 100)=11` meaning that we would need to do 11 "fan out" processes (so there are 10 "fan outs" of 100 records and an 11th of the remaining 7 records). The test shows that it will only process the first 1000 in this case, with 7 remaining records to be processed in the next iteration of the etl

src/layers/domain/repository/repository/v2.py
jaklinger: yeah if two or more processes fail at the same time it means they won't all restart at exactly the same time

jaklinger: jitter retrying is a strategy from my web scraping days

src/etl/sds/worker/transform_bulk/tests/test_transform_bulk_worker.py
Rohoolio: Ahhh gotcha!

Rohoolio (approved):

PR #314, opened 2024-09-16
feature/PI-495-generating_product_ids

megan-bower4:


src/layers/domain/core/cpm_system_id/tests/test_cpm_system_id_v1.py
```diff
@@ -250,16 +262,47 @@ def test_asid_generation_increment():
...
+
+@pytest.mark.parametrize(
+    "invalid_key",
+    [
+        "ABC000124",  # Missing 'P.' and hyphen
+        "P-ABC-123",  # Extra hyphen
+        "P.ABC123",  # Missing hyphen
+        "P.ABC-1234",  # Too many digits
+        "P.ABC-12",  # Too few digits
+        "P.AB1-123",  # ODS code contains a digit
```
Rohoolio, line 301: Slightly confusing comments here as there is no ODS code involved and the the sections arent specifically digit or not - they are alphanumeric with a reduced choice (due to confusing characters) so worth updating these :)

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -1,86 +1,139 @@
...
+    @classmethod
+    def _get_key_name(cls) -> CpmSystemIdType:
+        return CpmSystemIdType.PRODUCTID
+
+    @classmethod
+    def _format_key(cls) -> str:
+        """Generate the product ID as a random string in the format 'P.XXX-XXX'."""
+        PART_LENGTH = 3
+        N_PARTS = 2
+        PRODUCT_ID_VALID_CHARS = "ACDEFGHJKLMNPRTUVWXY34679"
```
Rohoolio, line 123: You have this in two separate methods on the class - might be worth pulling it out as a productID class variable? Especially since its a specific collection of characters its something that might be faffy to track if its in more than one place

src/layers/domain/core/cpm_system_id/tests/test_cpm_system_id_v1.py
megan-bower4: Good spot, some of these are actually wrong anyway! They use invalid characters so would always fail regardless 🤦‍♀️

jameslinnell (dismissed):

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -1,86 +1,124 @@
...
     def create(cls, current_id=None):
-        # Initialize current_id with a default if not provided
-        current_id = current_id["latest"] if current_id is not None else 200000099999
-        return super().create(key_name="ASIDNUMBER", current_id=current_id)
+        current_id = current_id.get("latest") if current_id else 200000099999
+        instance = super().create(
+            current_id=current_id, key_name=CpmSystemIdType.ASIDNUMBER
+        )
+        instance.latest = current_id + 1
+        instance.latest_id = cls._format_key(
```
jameslinnell, line 50: instance.latest = instance.latest_number
instance.current_id = instance.current_number
instance.latest_id = instance.id

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -1,86 +1,124 @@
...
 
-class CpmSystemId(BaseModel):
+class CpmSystemIdType(StrEnum):
+    ASIDNUMBER = auto()
+    PARTYKEYNUMBER = auto()
+    PRODUCTID = auto()
+
+
+class CpmSystemId(BaseModel, ABC):
     key_name: str
```
jameslinnell, line 18: Optional ???

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -1,86 +1,124 @@
...
-    latest: int
-    current_number: int
     latest_id: str
```
jameslinnell, line 19: Optional ???

megan-bower4: I've kept these in in the end, as although they aren't always used in the create method, they are always attributes needed on all the classes, and get populated during the process.

PR #340, opened 2024-10-02
[Feature/PI-498] Search for cpm products

jameslinnell:


src/api/tests/feature_tests/features/searchCpmProduct.failure.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Search CPM Products - failures scenarios
```
megan-bower4, line 1: Will there be any failure scenarios for missing parameters or extra parameters?

src/layers/domain/core/ods_organisation/tests/test_ods_organisation_v3.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.parametrize(
+    "ods_code",
+    [
+        "ABCDEFGHIJ",
+        "!@£$%",
+    ],
+)
+def test__id_must_be_ods_code(ods_code: str):
```
megan-bower4, line 24: Super minor but maybe call test__id_must_be_valid_ods_code or test__invalid_ods_code

src/layers/domain/repository/cpm_product_repository/tests/test_product_repository.py
```diff
@@ -0,0 +1,115 @@
...
+
+def _create_product_team(
+    seed: int = 1, name: str = "product-team-name", ods_code: str = "ABC"
+):
+    org = Root.create_ods_organisation(ods_code=ods_code)
+    return org.create_product_team(id=consistent_uuid(seed), name=name)
+
+
+@pytest.mark.integration
+def test__query_products_by_team():
```
megan-bower4, line 22: Maybe name test__query_products_by_product_team to be consistent with language, and makes it super easy to find the test for that function

src/layers/domain/repository/cpm_product_repository/tests/v3/test_cpm_product_repository_v3.py
```diff
@@ -105,3 +106,120 @@ def test__cpm_product_repository__product_does_not_exist_local():
         )
         with pytest.raises(ItemNotFound):
             repo.read(product_team_id=product_team_id, product_id=product_id)
+
+
+@pytest.mark.integration
+def test__query_products_by_team():
```
megan-bower4, line 112: What's the difference between these tests and the ones in the file above?

jameslinnell: Ah that other file shouldn't be there. Had it before I rebased yours

src/api/tests/feature_tests/features/searchCpmProduct.failure.feature
jameslinnell: Any other path params would cause a 404
I can see a future where we might want to pass in query params, however currently we don't parse any incoming query params so they would just be ignored. And it wouldn't fail. 
Same for body

megan-bower4: Ah yeah, I was getting confused between query and path params

megan-bower4 (approved):

PR #350, opened 2024-10-09
[Feature/PI-533] Create a productteam

jameslinnell:


src/api/createProductTeam/src/v1/steps.py
```diff
@@ -15,33 +20,51 @@ def parse_event_body(data, cache) -> dict:
     return event.json_body if event.body else {}
 
 
-def create_product_team(data, cache) -> ProductTeam:
+def parse_incoming_product_team(data, cache) -> CreateProductTeamIncomingParams:
```
jaklinger, line 23: rather than try/except, suggest to use the `@mark_validation_error_as_incoming`

src/layers/domain/core/ods_organisation/v3.py
```diff
@@ -15,8 +13,8 @@ class OdsOrganisation(AggregateRoot):
 
     ods_code: str = Field(regex=ODS_CODE_REGEX)
 
-    def create_product_team(self, id: UUID, name: str) -> ProductTeam:
-        product_team = ProductTeam(id=id, name=name, ods_code=self.ods_code)
+    def create_product_team(self, name: str = "", keys: list = []) -> ProductTeam:
```
jaklinger, line 16: 2 comments on the defaults:

* I don't think we should allow default `name` values
* For default lists you need to do `keys: list = None` and then in the function set the default (for info [see this](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument))

so suggestion would be

```suggestion
    def create_product_team(self, name: str, keys: list = None) -> ProductTeam:
        keys = keys or []
```

src/layers/domain/core/product_team/v3.py
```diff
@@ -22,9 +31,23 @@ class ProductTeam(AggregateRoot):
...
+    deleted_on: datetime = Field(default=None)
+    keys: list[ProductTeamKey] = Field(default_factory=list)
+
+    @root_validator(pre=True)
+    def set_id(cls, values):
+        if values.get("id") is None:
+            ods_code = values.get("ods_code")
+            if ods_code:
+                product_team_id = ProductTeamId.create(ods_code=ods_code)
+                values["id"] = product_team_id.id
```
jaklinger, line 49: maybe just to tidy up the indenting:

```suggestion
    def set_id(cls, values):
        ods_code = values.get("ods_code")
        if ods_code and not values.get("id"):
            values["id"] = ProductTeamId.create(ods_code=ods_code)
```

src/layers/domain/response/render_response.py
```diff
@@ -33,19 +36,31 @@ def render_response[
...
+        # Implicit failure from all other Exceptions
+        outcome = ErrorResponse.from_exception(exception=response).dict()
+        http_status = http_status_from_exception(exception=response)
     else:
-        # Implicit success (e.g. SEARCH, READ operations)
-        http_status = HTTPStatus.OK
-        outcome = response
+        if isinstance(response, tuple):
+            http_status = response[0]
+            outcome = response[1]
```
jaklinger, line 57: To be explicit (meaning we're also requiring there to be exactly do items in the tuple)

```suggestion
            http_status, outcome = response
```

src/layers/domain/response/render_response.py
```diff
@@ -33,19 +36,31 @@ def render_response[
...
         response = validate_exception(response)
 
-    if isinstance(response, HTTPStatus):
+    if response == HTTPStatus.NO_CONTENT:
+        http_status = response
+        outcome = None
+    elif isinstance(response, HTTPStatus):
         # Explicit success (e.g. CREATE, DELETE, UPDATE operations)
         http_status = response
         outcome = operation_outcome_ok(id=id, http_status=http_status)
```
jaklinger, line 45: is this block needed anymore?

pyproject.toml
```diff
@@ -55,6 +55,7 @@ boto3-stubs = { extras = [
 ], version = "^1.34.37" }
 pytest-timeout = "^2.2.0"
 lz4 = "^4.3.3"
+gevent = "<24.10.2"
```
jaklinger, line 58: the current release has built ok without this now, so i think good to remove

jaklinger (approved):

PR #383, opened 2024-10-30
[feature/PI-558] Add a search Results wrapper

jameslinnell:


src/layers/domain/response/response_models.py
```diff
@@ -0,0 +1,13 @@
...
+
+T = TypeVar("T", bound=AggregateRoot)
+
+
+class SearchResponse(Generic[T], AggregateRoot):
+    results: List[T]
+
+    @classmethod
+    def from_models(cls, models: List[T]) -> "SearchResponse[T]":
+        return cls(results=models)
```
jaklinger, line 13: from python 3.12 the typing syntax has dropped a lot of the need for imports

```suggestion
from domain.core.aggregate_root import AggregateRoot

class SearchResponse[T](AggregateRoot):
    results: list[T]

    @classmethod
    def from_models(cls, models: list[T]) -> "SearchResponse[T]":
        return cls(results=models)
```

src/layers/domain/response/response_models.py
```diff
@@ -0,0 +1,13 @@
...
+from domain.core.aggregate_root import AggregateRoot
+
+T = TypeVar("T", bound=AggregateRoot)
+
+
+class SearchResponse(Generic[T], AggregateRoot):
+    results: List[T]
+
+    @classmethod
+    def from_models(cls, models: List[T]) -> "SearchResponse[T]":
```
jaklinger, line 12: what's the advantage of the method `from_models` over the standard constructor `SearchResponse(results=models)`

megan-bower4 (approved):

PR #398, opened 2024-11-07
Feature/pi 581 create mhs device with drd

megan-bower4:


src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -0,0 +1,198 @@
...
+from domain.repository.questionnaire_repository.v2 import QuestionnaireRepository
+from domain.repository.questionnaire_repository.v2.questionnaires import (
+    QuestionnaireInstance,
+)
+from domain.request_models.v1 import CpmProductPathParams, CreateMhsDeviceIncomingParams
+from domain.response.validation_errors import mark_validation_errors_as_inbound
+
+
+@mark_validation_errors_as_inbound
+def parse_mhs_device_payload(data, cache) -> Device:
```
jameslinnell, line 36: Not returning a Device.

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -26,6 +43,64 @@ def parse_mhs_device_payload(data, cache) -> Device:
...
+        table_name=cache["DYNAMODB_TABLE"], dynamodb_client=cache["DYNAMODB_CLIENT"]
+    )
+    mhs_device = device_repo.search(
+        product_team_id=product_team.id, product_id=product.id
+    )
+    for device in mhs_device:
+        if device.name == "Product-MHS":
+            raise ConfigurationError(
+                f"There is already an exisitng MHS Device for this product, with device_id: {mhs_device[0].id}"
+            )
```
jaklinger, line 76: Suggest to 
* fix up the naming
* compress the for/if to one indent
* use a const for MHS device name (will need to be used by ETL)

```suggestion
    devices = device_repo.search(
        product_team_id=product_team.id, product_id=product.id
    )
    if any(device.name == MHS_DEVICE_NAME for device in devices)
        raise ConfigurationError(
            f"There is already an existing MHS Device for this product"
         ) 
```

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -26,6 +43,64 @@ def parse_mhs_device_payload(data, cache) -> Device:
...
+
+
+def read_device_reference_data(data, cache) -> DeviceReferenceData:
+    path_params: CpmProductPathParams = data[parse_path_params]
+    drd_repo = DeviceReferenceDataRepository(
+        table_name=cache["DYNAMODB_TABLE"], dynamodb_client=cache["DYNAMODB_CLIENT"]
+    )
+    device_reference_data = drd_repo.search(
```
jaklinger, line 84: rename to be plural

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -36,7 +111,7 @@ def validate_spine_mhs_questionnaire_response(data, cache) -> QuestionnaireRespo
     questionnaire_responses = payload.questionnaire_responses
 
     # Ensure there's a questionnaire named 'spine_mhs' in the responses
-    if QuestionnaireInstance.SPINE_MHS not in questionnaire_responses:
+    if "spine_mhs" not in questionnaire_responses:
```
jaklinger, line 114: 2 questions...

* why switch from using the constant? 
* i think maybe we got the logic slightly wrong the first time we wrote this - should it not be:

```suggestion
    if questionnaire_responses.keys() != {QuestionnaireInstance.SPINE_MHS}:
```

since we wouldn't want to allow any other questionnaire to be sent as well

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -64,32 +139,63 @@ def create_mhs_device(data, cache) -> Device:
...
+
+def create_interaction_id_keys(data, cache) -> DeviceKeyAddedEvent:
+    mhs_device: Device = data[create_mhs_device]
+    party_key = data[get_party_key]
+    drd: DeviceReferenceData = data[read_device_reference_data]
+    interaction_ids = []
+
+    # Extract Interaction IDs from questionnaire responses
+    questionnaire_responses = drd.questionnaire_responses.get(
+        "spine_mhs_message_sets/1", []
```
jaklinger, line 155: use `QuestionnaireInstance` constants

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -64,32 +139,63 @@ def create_mhs_device(data, cache) -> Device:
...
+    )
+    for response in questionnaire_responses:
+        interaction_ids.append(response.data.get("Interaction ID"))
```
jaklinger, line 158: I think probably better to pull the field `"Interaction ID"` out of the json schema as we're very likely to rename these fields

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -64,32 +139,63 @@ def create_mhs_device(data, cache) -> Device:
...
 
 
-def create_party_key_tag(data, cache):
+def create_party_key_tag(data, cache) -> DeviceTagAddedEvent:
     mhs_device: Device = data[create_mhs_device]
-    mhs_device.add_tag(party_key=data[get_party_key])
+    return mhs_device.add_tag(party_key=data[get_party_key])
+
+
+def create_interaction_id_keys(data, cache) -> DeviceKeyAddedEvent:
```
jaklinger, line 147: I don't think the MHS Device is supposed to have interaction id keys is it? I don't think they're necessarily unique per device. One to bring up over a chat with the team

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -64,32 +139,63 @@ def create_mhs_device(data, cache) -> Device:
...
+        "spine_mhs_message_sets/1", []
+    )
+    for response in questionnaire_responses:
+        interaction_ids.append(response.data.get("Interaction ID"))
+
+    # Use the extracted IDs to create keys for the device
+    for id in interaction_ids:
+        mhs_device.add_key(
+            key_type=DeviceKeyType.INTERACTION_ID, key_value=f"{party_key}:{id}"
+        )
```
jaklinger, line 164: i'm not sure it makes sense to prefix this with party key?

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -64,32 +139,63 @@ def create_mhs_device(data, cache) -> Device:
     return product.create_device(**device_payload)
 
 
-def create_party_key_tag(data, cache):
+def create_party_key_tag(data, cache) -> DeviceTagAddedEvent:
```
jaklinger, line 142: party key is supposed to be key rather than a tag, but whilst I'm here... we will need to add a bunch of tags to the device but perhaps best waiting until the bulk etl is done as then the code for doing that will be shareable (I'll write a ticket)

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -26,6 +43,64 @@ def parse_mhs_device_payload(data, cache) -> Device:
...
+    )
+
+    party_key: str = data[get_party_key]
+    mhs_message_set_drd_name = f"{party_key} - MHS Message Set"
+
+    device_reference_data = [
+        drd for drd in device_reference_data if drd.name == mhs_message_set_drd_name
+    ]
+
+    if len(device_reference_data) != 1:
```
jameslinnell, line 96: should this not be 1 or more? so 

`if not device_reference_data:`

This may be correct but I'm just going off what the first Scenario says

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
jaklinger: actually, now I'm thinking (I'm also using this constant) it would be best to have it as a constant, and have a test that checks that this constant matches the json schema

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
jameslinnell: Ok, maybe I'm wrong here. So it's one message set, but it can be multiple questionnaire responses?

megan-bower4: Yeah think the AC was just a general one, the actual task says

_if not exactly one Message Set DRD found then raise 5XX AccreditedSystemFatalError (shouldn't be possible)
returns the one Message Set DeviceReferenceData_

Because we only have one MHS per product, but you can have multiple AS's

megan-bower4: And again think there was some copy pasting overlap as the Error is MHS specific not Accredited System, and I also kept it as a 400 Validation error as that made more sense to me and would be more useful to the user I think

jameslinnell:
Don't forget the smoke tests! :)

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
megan-bower4: Added this logic into the incoming params to be flagged at the initial stage

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
megan-bower4: Changed to create_cpa_id_key for clarity - this will be unique per Spine_mhs_message_set response

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
megan-bower4: Party keys are tags for devices, keys at product level

jameslinnell (approved):

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -8,50 +8,91 @@
...
 
 
 @mark_validation_errors_as_inbound
-def parse_mhs_device_payload(data, cache) -> Device:
+def parse_mhs_device_payload(data, cache) -> CreateMhsDeviceIncomingParams:
     payload: dict = data[parse_event_body]
     return CreateMhsDeviceIncomingParams(**payload)
 
 
+def check_for_existing_mhs(data, cache):
```
jameslinnell, line 42: Still not sure about this. I don't see it on the diagram or in the ACs that it can only have one MHS_DEVICE.

PR #401, opened 2024-11-08
[feature/PI-601] Set branch ref to main if branch has been deleted on workspace delete

jameslinnell:


.github/workflows/on-pr-close.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ env.BRANCH_NAME }}
+          ref: ${{ env.BRANCH_NAME || 'main' }}
```
jameslinnell, line 31: So because the branch gets deleted it needs to run the build in "main". If it's just a regular close of PR this will remain on the branch name.

.github/workflows/on-pr-close.yml
```diff
@@ -41,7 +41,7 @@ jobs:
...
```
jameslinnell, line 44: So because the branch gets deleted it needs to run the build in "main". If it's just a regular close of PR this will remain on the branch name.

.github/workflows/on-pr-close.yml
jameslinnell: So BRANCH_NAME will have a value. As it's coming from the pull request. For definite. That's how `destroy--redundant-workspaces BRANCH_NAME=origin/${{ env.BRANCH_NAME }}` will still work as in this script the branch name is used to determine the workspace name. 

But the point is if the branch doesn't exist as a branch then it will fallback to main.

Apparently this works. But I'm more than happy to concede there is an element of doubt.
There is someone asking for it to be a feature https://github.com/actions/checkout/issues/512 back in 2021 and at the bottom there is someone showing a different option.

jaklinger (approved):

PR #409, opened 2024-11-15
[PI-528] collapse versions and deactivate etl tests

jaklinger:


src/layers/sds/cpm_translation/modify_key.py
```diff
@@ -1,186 +1,186 @@
...
-        key_value=new_scoped_party_key,
-    )
-    set_device_tags(
-        device=new_device,
-        data=questionnaire_response.flat_answers,
-        model=SearchSDSEndpointQueryParams,
-    )
-
-    yield new_device
+# from dataclasses import astuple
```
jameslinnell, line 1: Comments ???

jaklinger: ah this file is getting deleted soon and is getting used anywhere but could be useful for reference when rewriting the changelog

PR #418, opened 2024-11-19
[PI-605] EPR Bulk Create

jaklinger:


src/layers/sds/epr/bulk_create/creators.py
```diff
@@ -0,0 +1,106 @@
...
+def create_as_device(
+    product: CpmProduct,
+    party_key: str,
+    asid: str,
+    as_device_data: QuestionnaireResponse,
+    message_sets_id: str,
+    additional_interactions_id: str,
+    as_tags: list[dict],
+) -> Device:
+    # TODO: need to add asid keys to EPR create APIs!
```
megan-bower4, line 91: Is this captured in another ticket?

src/layers/sds/epr/constants.py
megan-bower4, line 1: Ooh I like this a lot! A lot more organised than what I did :)

src/layers/sds/epr/bulk_create/creators.py
jaklinger: this is being done by James now, removing comment!

PR #425, opened 2024-11-26
[feature/PI-650] Read as device

jameslinnell:


src/api/readDevice/src/v1/steps.py
```diff
@@ -74,14 +74,30 @@ def read_device_reference_data(data, cache) -> list[DeviceReferenceData]:
...
+    for key, filters in device.device_reference_data.items():
+        for drd in device_reference_data_list:
+            if str(drd.id) == key:
+                for responses in drd.questionnaire_responses.values():
+                    for qresponse in responses:
+                        if "*" not in filters:
+                            qresponse.data = {
+                                filter_key: qresponse.data[filter_key]
+                                for filter_key in filters
+                            }
```
megan-bower4, line 93: This is very heavily nested - although I've tried to look at it but not sure how to make it any simpler either, unless some of the logic is pulled out into a separate function for readability at least? Don't know if Joel would have any magic python ideas for this? He's improved a lot of my if statements/for loops before :)

jameslinnell: Yeah, it's horrible and it took me many attempts to write it. I guess we can split into functions.

jaklinger: I think the behaviour isn't quite right, as it should be able to deal with full paths to fields 

for the step I'd do something like

```python
def filter_device_reference_data(data, cache) -> list[QuestionnaireResponse]:
    device: Device = data["read_device"]
    device_reference_data_list: list[DeviceReferenceData] = data[
        "read_device_reference_data"
    ]

    filtered_questionnaire_responses = []
    for drd in device_reference_data_list:
        filters = device.device_reference_data[str(drd.id)]
        filtered_questionnaire_responses.append(
            {
                qid: [
                    filter_by_jsonpath(data=qr.data, filters=filters)
                    for qr in questionnaire_responses
                ]
                for qid, questionnaire_responses in drd.questionnaire_responses.items()
            }
        )
    return filtered_questionnaire_responses
```

with a filter function like

```python
def filter_by_jsonpath(data: dict | list[dict], jsonpath: str) -> dict | list[dict]: ...
```

that satisfies tests like

```python
sample = {
    "some": [{"deeply": {"nested": "field"}}, {"other": {"nested": "field"}}],
    "another": [{"deeply": {"nested": "field"}}, {"other": {"nested": "field"}}],
}

@pytest.mark.parametrize(
    ["path", "expected"],
    [
        (
            "*",
            sample,
        ),
        (
            "*.deeply.*",
            {
                "some": [{"deeply": {"nested": "field"}}],
                "another": [{"deeply": {"nested": "field"}}],
            },
        ),
        (
            "some.deeply.nested",
            {
                "some": [{"deeply": {"nested": "field"}}],
            },
        ),
    ],
)
def test_filter_by_jsonpath(path, expected):
    assert filter_by_jsonpath(data=sample, jsonpath=path) == expected
```

megan-bower4 (approved):

src/api/tests/feature_tests/features/createDeviceAccreditedSystem.success.feature
```diff
@@ -17,6 +17,7 @@ Feature: Create AS Device - success scenarios
       | path | value            |
       | name | My Great Product |
     And I note the response field "$.id" as "product_id"
+    And I note the response field "$.keys.0.key_value" as "party_key"
```
megan-bower4, line 20: I don't think this value is used/referenced anywhere?

megan-bower4 (approved):

PR #429, opened 2024-11-29
[feature/PI-631] Generate Product ids

jameslinnell:


scripts/administration/id_generator.py
megan-bower4, line 1: Super minor but probably call this file product_id_generator.py just so we don't get confused with other ids :)

scripts/administration/id_generator.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def bulk_generator(id_count=1, id_type="product"):
+    ids = open_file(id_type=id_type)
+    for _ in range(id_count):
+        saved = False
+        while not saved:
+            generated_id = generator()
+            saved = save_to_set(generated_id, ids)
+        ids = saved
```
jaklinger, line 49: maybe to make more concise:

```suggestion
    initial_len = len(ids)
    while len(ids) - initial_len < id_count:
        ids.add(generator())
```

src/layers/domain/core/cpm_system_id/tests/test_cpm_system_id_v1.py
```diff
@@ -1,5 +1,21 @@
...
+from event.json import json_load
+
+PATH_TO_CPM_SYSTEM_IDS = Path(__file__).parent.parent
+PRODUCT_IDS_GENERATED_FILE = f"{PATH_TO_CPM_SYSTEM_IDS}/generated_ids/product_ids.json"
+generated_product_ids = set()
+
+
+@pytest.fixture(scope="module", autouse=True)
+def _get_generated_ids():
+    global generated_product_ids
```
jaklinger, line 15:
```suggestion
generated_product_ids = set()


@pytest.fixture(scope="module")
def generated_product_ids():
```

and then use the `generated_product_ids` fixture in the arg of the required test?

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -105,19 +111,29 @@ def latest_number(self):
...
+        existing_ids = cls.load_existing_ids()
+        while True:
+            """No current_id needed, key is generated randomly."""
+            rng = random.Random(datetime.now().timestamp())
+            product_id = "-".join(
+                "".join(rng.choices(PRODUCT_ID_VALID_CHARS, k=PRODUCT_ID_PART_LENGTH))
+                for _ in range(PRODUCT_ID_NUMBER_OF_PARTS)
+            )
+            if f"P.{product_id}" not in existing_ids:
+                return cls(__root__=f"P.{product_id}")
```
jaklinger, line 123: to minimise the change, perhaps instead could just change the `return` line of the function to:

```
        if f"P.{product_id}" in cls.load_existing_ids():
            return cls.create()
        return cls(__root__=f"P.{product_id}")
```

scripts/administration/id_generator.py
jameslinnell: It would if we wanted in the future be able to generate other ids, it's just that by default it generates product_ids. So it is generic

src/layers/domain/core/cpm_system_id/v1.py
jameslinnell: I did it this way so that the file is called once. In your suggestion the file would get called on each recursive flow. Seems

src/layers/domain/core/cpm_system_id/tests/test_cpm_system_id_v1.py
jameslinnell: Either way the fixture is called once, but ok.

src/layers/domain/core/cpm_system_id/v1.py
jaklinger: i'd cache the load under the hood to avoid loading the file twice, so e.g.

```python
@cache
def _load_existing_ids():
   # load the ids

...

class CantRememberTheNameOfTheClass:
   ...

   @classmethod
   def load_existing_ids(cls):
       return _load_existing_ids()

   ...
```

megan-bower4 (approved):

PR #435, opened 2024-12-03
Feature/pi 645 update spine mhs message sets questionnaire

megan-bower4:


src/layers/domain/repository/questionnaire_repository/v1/questionnaires/field_generation.py
megan-bower4, line 1: Any suggestions for where this should live? Currently causes a circular import error here due to needing QuestionnaireInstance & being used by both the core domain and questionnaire repository

src/layers/domain/api/common_steps/questionnaire_response_validation.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+
+
+def process_and_validate_questionnaire_response(
+    questionnaire: Questionnaire,
+    questionnaire_response: dict,
+    party_key: str,
+    instance: QuestionnaireInstance,
+) -> QuestionnaireResponse:
+    # Check for user-provided fields
```
jaklinger, line 28: just a comment to say i don't think you need your comments :)

jaklinger (approved):

src/layers/sds/domain/nhs_mhs.py
```diff
@@ -47,19 +47,20 @@ class NhsMhs(SdsBaseModel):
     nhs_mhs_duplicate_elimination: Optional[YesNoAnswer] = Field(
         alias="nhsmhsduplicateelimination"
     )
+    binding: Optional[str] = Field(alias="binding")
```
megan-bower4, line 50: Potentially don't ask for binding and always combine https:// with fqdn when generating the endpoint/address

PR #436, opened 2024-12-04
[feature/PI-695] Add a custom flag to behave clear_screen=true

jameslinnell:


src/api/tests/feature_tests/environment.py
```diff
@@ -114,6 +118,9 @@ def after_scenario(context: Context, scenario: Scenario):
 def after_feature(context: Context, feature: Feature):
     if context.postman_feature:
         context.postman_collection.item.append(context.postman_feature)
+    if context.clear_screen:
+        click.clear()
+        time.sleep(0.5)
```
megan-bower4, line 123: Why do you need a sleep? It's too quick? 💨

megan-bower4 (approved):

src/api/tests/feature_tests/environment.py
jameslinnell: yeah, it's just to allow the refresh to happen before the next feature kicks off. It's 500ms so barely noticeable

PR #441, opened 2024-12-11
[feature/PI-652] Add a perception of an Environment layer

jameslinnell:


src/layers/domain/api/common_steps/sub_product.py
jaklinger, line 1: why `sub_product`?

src/api/createDevice/src/v1/steps.py
```diff
@@ -22,7 +23,8 @@ def parse_device_payload(data, cache) -> CreateDeviceIncomingParams:
 def create_device(data, cache) -> Device:
     product: CpmProduct = data[read_product]
     payload: CreateDeviceIncomingParams = data[parse_device_payload]
-    return product.create_device(**payload.dict())
+    environment = data[read_environment]
+    return product.create_device(env=environment, **payload.dict())
```
jaklinger, line 27: I thought it would work as `environment.create_product` and then `product.create_device` (and environment would be implicitly passed in)

src/api/createDeviceMessageHandlingSystem/src/v1/steps.py
```diff
@@ -116,11 +123,15 @@ def create_mhs_device(data, cache) -> Device:
     product: CpmProduct = data[read_product]
     party_key: str = data[get_party_key]
     payload: CreateMhsDeviceIncomingParams = data[parse_mhs_device_payload]
+    environment: Environment = data[read_environment]
+    # payload.__dict__["env"] = data[read_environment]
```
jaklinger, line 127: comment to remove

src/api/createDeviceReferenceData/tests/test_index.py
```diff
@@ -71,6 +72,7 @@ def test_index() -> None:
                 "pathParameters": {
                     "product_team_id": str(product.product_team_id),
                     "product_id": str(product.id),
+                    "env": Environment.DEV,
```
jaklinger, line 75: better to name `environment` rather than `env`? `env` to me sounds like environment variables

src/layers/domain/core/cpm_product/v1.py
```diff
@@ -75,6 +75,7 @@ class CpmProduct(AggregateRoot):
     def create_device(
         self,
         name: str,
+        env: Environment,
```
jaklinger, line 78: better to name `environment`?

src/layers/domain/core/enum.py
```diff
@@ -4,3 +4,14 @@
...
+
+class Environment(StrEnum):
+    DEV = auto()
+    QA = auto()
+    REF = auto()
+    INT = auto()
+    PROD = auto()
+
+    # def __str__(self) -> str:
+    #     return self.name.upper()
```
jaklinger, line 17: remove comment

src/api/createDevice/src/v1/steps.py
jameslinnell: But why environment.create_product? The environment is not on the product level. Also ticket just states it's an ENUM, not another layer of classes.

src/api/createDeviceReferenceData/tests/test_index.py
jameslinnell: yeah, probably.

PR #447, opened 2024-12-17
Billing alarms

jameslinnell:


infrastructure/terraform/per_account/dev/terraform.tf
```diff
@@ -13,5 +13,17 @@ terraform {
       source  = "hashicorp/aws"
       version = "~> 5.20.0"
     }
+    null = {
+      source  = "hashicorp/null"
+      version = "~> 3.0"
+    }
+    external = {
```
megan-bower4, line 20: Where are these 2 providers used?

megan-bower4 (approved):

infrastructure/terraform/per_account/dev/terraform.tf
jameslinnell: This was an issue that we had, it was complaining that they weren't there. Not sure if you were around for that conversation, basically the tf kept failing without spcifiying them

PR #450, opened 2024-12-18
Feature/pi 686 add an as

megan-bower4:


src/layers/sds/epr/creators.py
```diff
@@ -94,14 +94,12 @@ def create_as_device(
...
     message_sets_id: str,
     additional_interactions_id: str,
-    as_tags: list[dict],
 ) -> Device:
     as_device = product.create_device(
         name=EprNameTemplate.AS_DEVICE.format(party_key=party_key, asid=asid)
     )
     as_device.add_key(key_type=DeviceKeyType.ACCREDITED_SYSTEM_ID, key_value=asid)
     as_device.add_questionnaire_response(as_device_data)
-    as_device.add_tags(tags=as_tags)
```
jaklinger, line 104: don't think we want to remove tags do we?

src/layers/sds/epr/utils.py
```diff
@@ -32,7 +32,15 @@ def is_as_device(device: Device) -> bool:
...
-    return {qr.data[SdsFieldName.INTERACTION_ID] for qr in questionnaire_responses}
+
+    # Handle the case where questionnaire_responses is empty
+    if not questionnaire_responses:
+        return set()
+
+    return {
+        qr.data[SdsFieldName.INTERACTION_ID]
+        for qr in next(iter(questionnaire_responses), [])
+    }
```
jaklinger, line 46: fwiw an implicit way (and also preventing multiple `return`) would be

```python
    try:
        (questionnaire_responses,) = (
            message_sets_or_additional_interactions.questionnaire_responses.values()
        )
    except ValueError:
        interaction_ids = set()
    else:
        interaction_ids = {qr.data[SdsFieldName.INTERACTION_ID] for qr in questionnaire_responses}
    return interaction_ids
```

src/layers/sds/epr/creators.py
megan-bower4: Ah I was following the pattern for mhs, as they were removed there, and I thought they'd be captured in the tags ticket. But do AS Devices still need the tags of the sds fhir query params? And mhs have no tags then?

jaklinger: yeah AS tags are still needed for SDS FHIR API - but there are fewer of them, and MHS devices don't have tags any more
=

jaklinger (approved):

jaklinger (approved):

PR #468, opened 2025-01-16
[feature/PI-714] Add some new config to swagger

jameslinnell:


infrastructure/swagger/11_components--requestBodies.yaml
megan-bower4, line 1: On line 39 of this file it should be 
key_value: "FOOBAR"
to match the example response (or change to bar in the example, just to align them)

image

megan-bower4: Same for example product name on line 54 - change to 'Sample Product Team'

image

infrastructure/swagger/07_components--schemas--domain.yaml
megan-bower4, line 1: The example device reference data response isn't quite right here - the questionnaire response on line 314 should have an example, not key values in there.
E.g. ```spine_as_additional_interactions/1:
question: answer```
...

image

megan-bower4: Also needs updating for sample device response (about line 211 I think)

image

megan-bower4: And the get device and get device reference data examples need questionnaire_responses updating to the same

infrastructure/swagger/05_paths.yaml
```diff
@@ -509,6 +588,16 @@ paths:
     get:
       operationId: readQuestionnaire
       summary: Read a Questionnaire resource (GET)
+      description: |
+        Retrieve information about the format of a questionnaire to be answered. To be used in conjunction with creating Devices and DeviceReferenceData. **Please note: Although examples are given it's impossible to layout exactly what format a questionnaire will take as the questionnaire itself uses json_schema to allow custom formats.**
```
megan-bower4, line 592: Did we want to include a small description of what the fields on the questionnaire mean, e.g. what does system generated mean to the user? What does required mean? etc

infrastructure/swagger/07_components--schemas--domain.yaml
```diff
@@ -108,8 +108,40 @@ components:
...
+                      type: string
+                    system generated:
+                      type: boolean
+      example:
+        name: "questionnaire_1"
+        version: "1"
+        json_schema:
+          $schema: "http://json-schema.org/draft-07/schema#"
+          type: "object"
+          properties:
```
megan-bower4, line 138: Add the required block at the end of the questionnaire example,. as all our questionnaires have that block

infrastructure/swagger/05_paths.yaml
```diff
@@ -149,6 +183,10 @@ paths:
     delete:
       operationId: deleteproductendpoint
       summary: Delete a Product resource (DELETE)
+      description: |
+        - Delete a product using a product team ID and product ID.
+      tags:
+        - Core Deletion Endpoints
```
megan-bower4, line 189: There's no delete product operation successful example response

image

infrastructure/swagger/11_components--requestBodies.yaml
```diff
@@ -87,6 +87,9 @@ components:
                       description: Questionnaire Responses
             required:
               - questionnaire_responses
+          example:
+            questionnaire_responses:
+              { "spine_mhs_message_sets": [{ "question": "answer" }] }
```
megan-bower4, line 92: Include the versioning here so that they aren't confused when they see that? "spine_mhs_message_sets/1": ...

infrastructure/swagger/05_paths.yaml
jameslinnell: It think because there is not response other than the status code.

infrastructure/swagger/05_paths.yaml
jameslinnell: I've added System generated explanation but I think "required" is self explanatory?

infrastructure/swagger/07_components--schemas--domain.yaml
```diff
@@ -108,8 +108,40 @@ components:
     Questionnaire:
       type: object
       properties:
-        tbc:
+        name:
+          type: string
+        version:
```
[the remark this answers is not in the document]
jameslinnell: ok, fields on this file are responses, not requests. Maybe

infrastructure/swagger/07_components--schemas--domain.yaml
jameslinnell: This is the response, it matches what the feature tests show as an expected response.

infrastructure/swagger/11_components--requestBodies.yaml
jameslinnell: Line 54 is a product. The description was incorrect

infrastructure/swagger/07_components--schemas--domain.yaml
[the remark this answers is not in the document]
jameslinnell: Version and all headers should be "Required" if possible

infrastructure/swagger/07_components--schemas--domain.yaml
jameslinnell: Look at a json schema and  update to match "properties" and "required"

megan-bower4:
Some of the example values are still mismatched (e.g. names used in the request not matching the name shown in the repose) and some questionnaire_responses still have key_type and key_value instead of question and answer. 
However this is now de-prioritised, and we are getting value from the re-structuring only. The clean up will be covered in another ticket when we revisit our swagger for EPR.

megan-bower4 (dismissed):

PR #476, opened 2025-01-23
[feature/PI-744] Create a new ProductTeam for CPM path

jameslinnell:
Update tests and ProductTeam ID

Set Delete path

Fix swagger paths

### DIFF
--- a/src/api/createCpmProduct/tests/test_index.py
+++ b/src/api/createCpmProduct/tests/test_index.py
@@ -4,7 +4,7 @@ from unittest import mock
 
 import pytest
 from domain.core.root import Root
-from domain.repository.product_team_repository import ProductTeamRepository
+from domain.repository.product_team_epr_repository import ProductTeamRepository
 from event.json import json_loads
 
 from test_helpers.dynamodb import mock_table
--- a/src/api/createCpmProductForEpr/src/v1/steps.py
+++ b/src/api/createCpmProductForEpr/src/v1/steps.py
@@ -9,7 +9,7 @@ from domain.api.common_steps.create_product import (
 from domain.core.cpm_product import CpmProduct
 from domain.core.cpm_system_id import PartyKeyId
 from domain.core.product_key import ProductKeyType
-from domain.core.product_team import ProductTeam
+from domain.core.product_team_epr import ProductTeam
 from domain.repository.cpm_system_id_repository import CpmSystemIdRepository
 
 
--- a/src/api/createProductTeamEpr/src/v1/steps.py
+++ b/src/api/createProductTeamEpr/src/v1/steps.py
@@ -1,0 +1,67 @@
+from http import HTTPStatus
+
+from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
+from domain.core.product_team_epr import ProductTeam
+from domain.core.root import Root
+from domain.ods import validate_ods_code
+from domain.repository.product_team_epr_repository import ProductTeamRepository
+from domain.request_models import CreateProductTeamIncomingParams
+from domain.response.validation_errors import (
+    mark_json_decode_errors_as_inbound,
+    mark_validation_errors_as_inbound,
+)
+from event.step_chain import StepChain
+
+
+@mark_json_decode_errors_as_inbound
+def parse_event_body(data, cache) -> dict:
+    event = APIGatewayProxyEvent(data[StepChain.INIT])
+    return event.json_body if event.body else {}
+
+
+@mark_validation_errors_as_inbound
+def parse_incoming_product_team(data, cache) -> CreateProductTeamIncomingParams:
+    json_body = data[parse_event_body]
+    incoming_product_team = CreateProductTeamIncomingParams(**json_body)
+    return incoming_product_team
+
+
+def validate_product_team_ods_code(data, cache) -> None:
+    product_team_params: CreateProductTeamIncomingParams = data[
+        parse_incoming_product_team
+    ]
+    validate_ods_code(ods_code=product_team_params.ods_code)
+
+
+def create_product_team(data, cache) -> ProductTeam:
+    product_team_params: CreateProductTeamIncomingParams = data[
+        parse_incoming_product_team
+    ]
+    org = Root.create_ods_organisation(ods_code=product_team_params.ods_code)
+    product_team = org.create_product_team_epr(
+        **product_team_params.dict(exclude={"ods_code"})
+    )
+    return product_team
+
+
+def save_product_team(data, cache) -> dict:
+    product_team: ProductTeam = data[create_product_team]
+    product_team_repo = ProductTeamRepository(
+        table_name=cache["DYNAMODB_TABLE"], dynamodb_client=cache["DYNAMODB_CLIENT"]
+    )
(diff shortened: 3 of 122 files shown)


src/api/createProductTeamEpr/src/v1/steps.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@mark_json_decode_errors_as_inbound
+def parse_event_body(data, cache) -> dict:
+    event = APIGatewayProxyEvent(data[StepChain.INIT])
+    return event.json_body if event.body else {}
+
+
+@mark_validation_errors_as_inbound
+def parse_incoming_product_team(data, cache) -> CreateProductTeamIncomingParams:
```
megan-bower4, line 23: Do we want to duplicate the CreateProductTeamIncomingParams to have a specific Epr case? I know they both share the same for now, but can then delete all the epr bits without any side effects?

src/api/createCpmProduct/tests/test_index.py
```diff
@@ -4,7 +4,7 @@
 
 import pytest
 from domain.core.root import Root
-from domain.repository.product_team_repository import ProductTeamRepository
+from domain.repository.product_team_epr_repository import ProductTeamRepository
```
megan-bower4, line 7: Ah I assumed the class name would have changed also - so we have a ProductTeamRepository and ProductTeamEprRepository to make it very clear what is being used where? Without needing to trace the import back. But I guess that's just preference really

src/api/tests/smoke_tests/test_smoke.py
```diff
@@ -41,7 +41,18 @@ def _request(base_url: str, headers: dict, path: str, method: str):
             ],
         ],
         [
-            "/ProductTeam/123/Product",
+            "/ProductTeamEpr",
```
megan-bower4, line 44: Are there smoke test for the 'new' ProductTeam endpoints? They all looked to have been renamed to Epr

megan-bower4: None ignore me, it's literally above this

src/etl/sds/worker/bulk/tests/test_bulk_e2e.py
```diff
@@ -8,13 +8,13 @@
 from domain.core.device.v1 import Device
 from domain.core.device_reference_data.v1 import DeviceReferenceData
 from domain.core.enum import Environment
-from domain.core.product_team.v1 import ProductTeam
+from domain.core.product_team_epr import ProductTeam
```
megan-bower4, line 11: Does it matter that it's chopped of the .v1 on some imports but left it for others? As we only have the 1 version I think it defaults to that right - but if you do make any more commits then may as well have them consistent, if not then happy to leave it

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -142,10 +145,14 @@ def load_existing_ids(cls):
 
 class ProductTeamId(CpmSystemId):
     @classmethod
-    def create(cls, ods_code: str):
-        return cls(__root__=f"{ods_code}.{uuid4()}")
+    def create(cls, ods_code: str = None):
+        if ods_code:
+            return cls(__root__=f"{ods_code}.{uuid4()}")
+        return cls(__root__=str(uuid4()))
```
megan-bower4, line 151: Is this function shared between them both still? Is it worth separating to a ProductTeamEprID so that it can just be chopped out easily when needed?

src/layers/domain/core/cpm_system_id/v1.py
```diff
@@ -29,6 +29,9 @@
 PRODUCT_TEAM_ID_PATTERN = re.compile(
     r"^[a-zA-Z0-9]+\.([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12})$"
 )
+PRODUCT_TEAM_ID_PATTERN_WITHOUT_ODS = re.compile(
```
megan-bower4, line 32: This without ods if for the 'new' product ID product team right? So this should probably have the original name of PRODUCT_TEAM_ID_PATTERN and rename the other epr specific one to like PRODUCT_TEAM_ERP_ID_PATTERN - as the with or without ods may get confusing

src/layers/domain/repository/product_team_epr_repository/v1.py
```diff
@@ -0,0 +1,40 @@
+from attr import asdict
+from domain.core.product_team_epr import ProductTeam, ProductTeamCreatedEvent
+from domain.core.product_team_key import ProductTeamKey
+from domain.repository.keys import TableKey
+from domain.repository.repository import Repository
+
+
+class ProductTeamRepository(Repository[ProductTeam]):
```
megan-bower4, line 8: Ah yeah this is where I expected the class names to all include Epr for epr specific cases now?

src/api/createProductTeamEpr/src/v1/steps.py
jameslinnell: I kind of felt that shared stuff like this would be ok. The likelihood is that will will remove the "EPR" productTeam eventually so we can make any separations out at that point. We also share these kind of steps (regarding incoming param validation) across different entities as well so if we were to separate them we could probably argue we showed for everything.

src/api/createCpmProduct/tests/test_index.py
jameslinnell: just felt the class name can stay the same. They are only ever used separately so there's no confusion on the files themselves and i can think of only 1 file where I import both and then use aliases. To change ProductTeamRepository and ProductTeam to ProductTeamEprRepository and ProductTeamEpy would end up with a huge amount of changes as that would need to propergate through. This seems like the simplest, quickest solution.

src/etl/sds/worker/bulk/tests/test_bulk_e2e.py
jameslinnell: the v1 is a hangup from the versioning days. We could definitely go through and change that everywhere. but I think that's probably more for tech debt than this ticket.

src/layers/domain/core/cpm_system_id/v1.py
jameslinnell: again, just seemed logical to have a shared productTeamId class rather than split them out.

src/layers/domain/repository/product_team_epr_repository/v1.py
jameslinnell: See message above. Happy to discuss.

megan-bower4 (approved):

src/api/createCpmProductForEpr/src/v1/steps.py
```diff
@@ -9,7 +9,7 @@
 from domain.core.cpm_product import CpmProduct
 from domain.core.cpm_system_id import PartyKeyId
 from domain.core.product_key import ProductKeyType
-from domain.core.product_team import ProductTeam
+from domain.core.product_team_epr import ProductTeam
```
deepaattiguppe, line 12: test comment

PR #478, opened 2025-01-27
feature/PI-726-update_as_tags_on_add_an_as Update tags for any existi…

megan-bower4:
…ng as devices when new as device added


src/etl/sds/tests/changelog/conftest.py
```diff
@@ -9,6 +9,10 @@
 
 import boto3
 import pytest
+from domain.core.cpm_product.v1 import CpmProduct
```
jameslinnell, line 12: Anywhere you have CPMProduct and ProductTeam now need to change to EPRProduct and the EPR ProductTeam

jameslinnell (approved):

PR #499, opened 2025-02-17
feature/PI-754-search_product

megan-bower4:


infrastructure/swagger/05_paths.yaml
```diff
@@ -289,25 +289,67 @@ paths:
...
-    #     <<: *ApiGatewayIntegration
-    #     uri: ${method_searchCpmProduct}
-    #   security:
-    #     - ${authoriser_name}: []
-    #     - app-level0: []
+
+  /Product:
+    options:
+      operationId: searchproductcors
+      summary: Search Products (OPTIONS)
```
jameslinnell, line 296: need to add "tags"

jameslinnell:
Add smoke test.

infrastructure/swagger/05_paths.yaml
megan-bower4: We don't need it here otherwise options appears in the list, it's on line 334 so that just the /Product endpoint is in the list :)

jameslinnell: we do, as they end up in the "default" category. Check the UI ticket. I think Rowan just tagged them with the same tag.

src/layers/domain/api/sds/query.py
```diff
@@ -45,3 +45,31 @@ def check_filters(cls, values: dict):
 
     def get_non_null_params(self):
         return self.dict(exclude_none=True)
+
```
jameslinnell, line 48: Think this needs to go in it's own file outside of SDS as this folder and file will eventually get removed

infrastructure/swagger/05_paths.yaml
```diff
@@ -289,25 +295,69 @@ paths:
       security:
         - ${authoriser_name}: []
         - app-level0: []
-    # get:
```
jameslinnell, line 292: Just go!!!

jameslinnell (dismissed):

PR #512, opened 2025-02-21
feature/PI-794-disable_epr_s3_integratioon_tests

megan-bower4:
…on tests

### DIFF
--- a/src/layers/etl_utils/io/test/test_io_dump.py
+++ b/src/layers/etl_utils/io/test/test_io_dump.py
@@ -1,30 +1,30 @@
-from io import BytesIO
+# from io import BytesIO
 
-import pytest
-from etl_utils.io import pkl_dump_lz4, pkl_dumps_lz4, pkl_load_lz4
-from etl_utils.io.test.io_utils import pkl_loads_lz4
-from event.json import json_load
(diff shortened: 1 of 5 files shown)


src/layers/etl_utils/io/test/test_io_dump.py
```diff
@@ -1,14 +1,11 @@
...
 from etl_utils.io import pkl_dump_lz4, pkl_dumps_lz4, pkl_load_lz4
 from etl_utils.io.test.io_utils import pkl_loads_lz4
 from event.json import json_load
 
-from etl.sds.tests.constants import EtlTestDataPath
 
-
-@pytest.mark.s3(EtlTestDataPath.FULL_JSON)
+# @pytest.mark.s3(EtlTestDataPath.FULL_JSON) Uncomment this when archived
 def test_pkl_lz4(test_data_paths):
```
jameslinnell, line 9: How does this work? Do they now run as unit tests?

jameslinnell: Should we not comment out the whole test?

megan-bower4: Yep just realised that 🤦‍♀️

jameslinnell (approved):

PR #514, opened 2025-02-24
Feature/pi 817 dependabot

megan-bower4:
- poetry: update attrs requirement from ^24.2.0 to ^25.1.0
- poetry: update black requirement from ^24.1.1 to ^25.1.0


megan-bower4:
No fixes were needed so the original dependabot branches can be merged directly into a release branch.

PR #516, opened 2025-02-25
[feature/PI-834] Remove epr domain logic

jameslinnell:


src/layers/domain/core/product_team_key/v1.py
```diff
@@ -20,10 +22,32 @@ def pattern(self) -> re.Pattern:
...
+
+    @property
+    def parts(self):
+        return (self.key_type, self.key_value)
+
+    def __hash__(self):
+        return hash(self.parts)
+
+
+def validate_key(key_value: str, key_type: ProductTeamKeyType):
```
megan-bower4, line 46: Is this duplicated in src/layers/domain/core/cpm_system_id/v1.py? Should it be here only?

megan-bower4 (dismissed):

megan-bower4 (approved):

PR #524, opened 2025-03-04
[Feature/PI-832] add product team id alias

jameslinnell:


src/api/readCpmProduct/tests/test_index.py
```diff
@@ -67,14 +69,14 @@ def test_index(version):
...
+    assert response_body["cpm_product_team_id"] == product_team.id
     assert response_body["name"] == PRODUCT_NAME
     assert response_body["ods_code"] == ODS_CODE
     assert response_body["updated_on"] is None
     assert response_body["deleted_on"] is None
 
     # Assertions for fields that only need to be included
-    assert "product_team_id" in response_body
+    assert "cpm_product_team_id" in response_body
     assert "created_on" in response_body
```
Rohoolio, line 80: Not massive but maybe im confused, how can product team id be needed in both exact match and optional?

Rohoolio (approved):

src/api/readCpmProduct/tests/test_index.py
jameslinnell: TBH, not entirely sure the point of the second assertion. Seems to be a mistake. I guess it's not doing any harm.

PR #550, opened 2025-03-19
[feature/PI-815] Recreate a product Team

jameslinnell:


src/api/tests/feature_tests/features/deleteProductTeam.success.feature
```diff
@@ -39,11 +39,19 @@ Feature: Delete Product Team - success scenarios
...
-      | path             | value                                                    |
-      | errors.0.code    | RESOURCE_NOT_FOUND                                       |
-      | errors.0.message | Could not find ProductTeam for key ('<product_team_id>') |
+      | path             | value                                                             |
+      | errors.0.code    | RESOURCE_NOT_FOUND                                                |
+      | errors.0.message | Could not find ProductTeam for key ('${ note(product_team_id) }') |
+    And the response headers contain:
+      | name         | value            |
+      | Content-Type | application/json |
+    When I make a "GET" request with "default" headers to "ProductTeam/0a78ee8f-5bcf-4db1-9341-ef1d67248715"
```
megan-bower4, line 50: You shouldn't need to specifically do this is the examples block includes product_team_id and the alias/key value already

src/api/tests/feature_tests/features/deleteProductTeam.success.feature
```diff
@@ -52,3 +60,71 @@ Feature: Delete Product Team - success scenarios
...
+    Then I receive a status code "404" with body
+      | path             | value                                                                       |
+      | errors.0.code    | RESOURCE_NOT_FOUND                                                          |
+      | errors.0.message | Could not find ProductTeam for key ('0a78ee8f-5bcf-4db1-9341-ef1d67248715') |
+    And the response headers contain:
+      | name         | value            |
+      | Content-Type | application/json |
```
megan-bower4, line 130: If you make this a scenario outline with examples, you can interchangeably use the internal prod id and the external key id - will also test that you can delete it using the key as well then, otherwise will need a test for that too

src/layers/domain/repository/product_team_repository/v1.py
```diff
@@ -50,9 +52,29 @@ def handle_ProductTeamCreatedEvent(self, event: ProductTeamCreatedEvent):
...
+        timestamp = int(datetime.now().timestamp() * 1000)
+        product_team_key_create_transactions = [
+            self.create_index(
+                id=f"{key.key_value}-{timestamp}",
+                parent_key_parts=(f"{key.key_value}-{timestamp}",),
+                data=asdict(event),
+                root=False,
+                row_type=EntityType.PRODUCT_TEAM_ALIAS,
+            )
+            for key in keys
```
megan-bower4, line 68: Just trying to picture this - so for every key that was on the root product team, a new row will be created with the same keys or with no keys now?

megan-bower4 (approved):

PR #559, opened 2025-03-27
feature/PI-870-sonarcloud_stuff

Rohoolio:


src/api/tests/feature_tests/environment.py
```diff
@@ -82,22 +82,6 @@ def before_feature(context: Context, feature: Feature):
         name=feature.name,
         description=" ".join(feature.description),
     )
-    cpm_scenarios = [
```
megan-bower4, line 85: Do we need these for the tests to be picked up?

jameslinnell: Hmm, this used to be used for determining which database to use. but that must have been taken out in a previous ticket. This just looks to be a list that is unused now.

megan-bower4 (approved):

PR #575, opened 2025-04-24
[feature/PI-880] Allow dependabot branch names

jameslinnell:


scripts/workflow/check-branch-name.sh
```diff
@@ -6,14 +6,18 @@ current_branch=$(git rev-parse --abbrev-ref HEAD)
 # Define a regular expression pattern to match the desired branch naming conventions
 feature_pattern="^feature/PI-[0-9]+[-][A-Za-z0-9\_]*$"
 release_pattern="^release/[0-9]{4}-[0-9]{2}-[0-9]{2}(-[a-z])?$"
+dependabot_pattern="^dependabot/.*"
```
megan-bower4, line 9: What does the .* do? Why not just *?

jameslinnell: it allows any character, then multiple times

megan-bower4 (approved):

